PR #4, opened 2023-02-15
1 - Create the skeleton of the React SPA

### LINKED ISSUE #1: Create the skeleton of the React application

Even though this issue is created without having established the full scope of the SPA MVP first milestone beforehand, we already know that we need this first issue to establish the skeleton of the React application.

The goal is to create a primitive React SPA, empty in terms of Dataverse features, with a simple "Hello Dataverse" message. This will initialize the codebase for this project, which we will evolve in subsequent issues.

Dataverse frontend developers

## What inspired the request?
- Current work on SPA MVP Strategy
- MVP Strategy document: [link](https://docs.google.com/document/d/1vOSsIGGCbgal4f_lSp1yInQifBcmo5vZ-Z9pfkVDyG4)

N/A

## Any brand new behavior do you want to add to Dataverse Frontend?
React application setup with "Hello Dataverse" message

## Any related open or closed issues to this feature request?

- https://github.com/IQSS/dataverse/issues/9347

MellyGray:

This PR creates the React application skeleton for the frontend rarchitecture. It provides with a basic skeleton with all the needed configuration to start developing a feature/component for the Dataverse SPA. Additionally, React default component has been replaced by a HelloDataverse component.

Closes https://github.com/IQSS/dataverse-frontend/issues/1

#### Changes

- [React App with TypeScript](https://create-react-app.dev/docs/getting-started#creating-a-typescript-app)
- [Jest](https://jestjs.io/docs/getting-started) and [Testing Library ](https://testing-library.com/docs/react-testing-library/intro) configured for unit tests
- [Cypress](https://docs.cypress.io/guides/component-testing/react/quickstart) for e2e tests
- [i18next](https://react.i18next.com) English language configuration for internationalization
- [ESlint](https://github.com/jsx-eslint/eslint-plugin-react) and [Stylelint](https://stylelint.io/user-guide/get-started) for linting
- [Prettier](https://prettier.io/docs/en/install.html) formatter
- Workflows for lint and tests
- [React Router](https://reactrouter.com/en/main/routers/create-browser-router)
- [SASS](https://create-react-app.dev/docs/adding-a-sass-stylesheet/) and [Bootstrap](https://getbootstrap.com/docs/5.0/customize/sass/) for stylesheets
- Add HelloDataverse component

#### React Starter 

More common tools:

- [Vite](https://vitejs.dev). Uses different bundler for production and development. The development bundler is faster but more unstable. This option was discarded because it may be differences between dev and production environments because of this 2 different bundlers. 

- [Create React App](https://create-react-app.dev/docs/getting-started). Uses webpack which is slower but the most stable solution right now. Really easy set up with the integrated scripts. In the future React may change the bundler from webpack to another better solution but our code doesn't know anything about the bundler so it won't be affected. The code is not coupled to the bundler.

#### Testing
Jest with Testing Library chosen because it allows unit testing of the React components from the user perspective instead of directly testing the business implementation. This means that the unit tests are not modified after refactoring a component because the user should see the same behaviour. Some other libraries check directly the props or states which means that if you refactor a prop variable name then the unit test will stop working.

Cypress has great support for React and Testing Library. It is a widely used solution that works great.

#### i18next

i18next is already being used in the Dataverse project. It has been configured in the React skeleton because in the very first component there is going to be text to translate, so it is easier to have the i18n configuration from the beginning.

I also see an advantage in not having to see the text in the components because it allows you to see a structure instead of a component with a concrete use case, for example in HelloDataverse.tsx I know there is a translated title but I don't know what the specific text is:

`<h1 className={styles.title}>{t('title')}</h1>`

It is really recommended to see the video on [i18next Introduction](https://react.i18next.com)

#### ESlint, Stylelint and Prettier

A basic ESlint configuration has been added, we may want to add more rules in the future or change the existing ones but the configuration is already done to ensure the whole project is following the same linting rules.

It is recommended to run prettier formatted on save, see [IntelliJ](https://www.jetbrains.com/help/idea/prettier.html#ws_prettier_run_automatically_in_current_project) and [VSCode](https://www.alphr.com/use-prettier-vs-code/) configuration

#### React Router

From [React Router](https://reactrouter.com/en/main/start/overview):
> React Router enables "client side routing".
> 
> Client side routing allows your app to update the URL from a link click without making another request for another document from the server. Instead, your app can immediately render some new UI and make data requests with fetch to update the page with new information.

It has been configured since the beginning because it is going to be used with every new page.

#### SASS and Bootstrap
Bootstrap is being used in the current Dataverse frontend so it will be used in the SPA too to maintain the UI appearance.

SASS is a really useful CSS extension, it has nesting, mixins and inheritance among other features to help with the CSS maintenance. 

In React it can used like:

```
import styles from './HelloDataverse.module.scss'

<div className={styles.container}>
  <header>
    <h1 className={styles.title}>{t('title')}</h1>
  </header>
</div>
```

I really like that there is no need to think of class names, you can just call the class name `.container` in every component and there are not going to be conflicts between any other classes because  `.container` is automatically compiled to the class name `.HelloDataverse_container`. 

This also forces you to not reuse CSS classes between components. **Reuse the component, not the CSS class.**

#### Workflows
I added 2 basic workflows to check the linting and the tests. They are configured to run on push and for the moment they are not taking so long. Test (5min) and lint(2min) but if they start to take longer then they might be configured to run only on merge.

The test workflow runs both unit and e2e tests. We may want to split them at some point and for example run the unit tests on push and the e2e on merge.

#### Hello Dataverse component

HelloDataverse.tsx is just the default Create React Component but adapted to Dataverse to have an overview of what a component would look like and test some initial configurations such as the translation hook from i18n.

This component will be removed once the first feature is developed.

Install `node v19` and `npm v9`.

Run `npm install`

Run `npm run start` and go to http://localhost:3000 to see the application

Run `npm run test` for the unit tests

Run `npm run cy:run` for the e2e tests

[Hello Dataverse screen](https://design.penpot.app/#/workspace/66c3210d-ba6d-813c-8002-07b562df1f0f/30b13804-acfd-8090-8002-10b6686db339?page-id=30b13804-acfd-8090-8002-10b6686db33a) If you cannot access the screen please contact melinahernandez@theagilemonkeys.com 

Not sure, the Hello Dataverse component will be removed for the release, but the skeleton will remain.

## Additional documentation:

[Dataverse - SPA Development Strategy](https://docs.google.com/document/d/1vOSsIGGCbgal4f_lSp1yInQifBcmo5vZ-Z9pfkVDyG4/edit#heading=h.mv60u03m9o2a)


README.md
```diff
@@ -1 +1,36 @@
...
\ No newline at end of file
+# Dataverse Frontend
+
+[![Tests](https://github.com/IQSS/dataverse-frontend/actions/workflows/test.yml/badge.svg)](https://github.com/IQSS/dataverse-frontend/actions/workflows/test.yml)
+
+## Available Scripts
+
+In the project directory, you can run:
+
+### `npm start`
```
donsizemore, line 9: Please include `npm install` so absolute novices may follow along?

package.json
```diff
@@ -0,0 +1,80 @@
+{
+  "name": "dataverse-frontend",
+  "version": "0.1.0",
+  "private": true,
+  "dependencies": {
+    "@testing-library/jest-dom": "^5.16.5",
+    "@testing-library/react": "^13.4.0",
+    "@testing-library/user-event": "^13.5.0",
+    "@types/node": "^16.18.12",
```
pdurbin, line 9: Does this mean we should be using node 16.18.12? If so, I can't install it:

```
$ nvm install 16.18.12
Version '16.18.12' not found - try `nvm ls-remote` to browse available versions.
```

I tried with just `nvm install 16.18` and that worked, giving me node 16.18.1.

I see that in GitHub Actions there's a newer version of node. I'm making an assumption that I should use the version here in package.json but I'm not sure. 😅

src/sections/hello-dataverse/HelloDataverse.tsx
```diff
@@ -0,0 +1,26 @@
...
+export function HelloDataverse() {
+  const { t } = useTranslation('helloDataverse')
+
+  return (
+    <div className={styles.container}>
+      <header>
+        <h1 className={styles.title}>{t('title')}</h1>
+        <img src={logo} className={styles.logo} alt={t('altImage') ?? 'logo'} />
+        <p>
+          <Trans t={t} i18nKey="description" components={{ 1: <code /> }} />
```
pdurbin, line 14: This Trans syntax is a little rough but I'm glad to see that for a simple case like `{t('title')}` it looks nice. 😄 

Here I see you're substituting a `<code>` tag. Fancy. 🧐  And a little ugly. 😜

pdurbin (commented):
I don't have much to say except that this is looking great! 🎉 🎉 🎉 

Go go go!! 🚀 🚀 🚀 

(I did leave a few comments.) 😄

README.md
MellyGray: Sure! Thank you!

package.json
MellyGray: Ups, I forgot to add the node version to the package.json.

`@types/node` is the library for the TypeScript node types, so the version do not correspond to the node version.

I am currently using node v19 that is why in the Github Actions it appears as node 19, but node v16 would work as well.

I edited package.json to add the node and npm version

```
"engines" : {
    "npm" : ">=8.0.0",
    "node" : ">=16.0.0"
  },
```

The version used should be greater or equal than those but I'm using node v19 and npm v9. Maybe I should add it to the README too

src/sections/hello-dataverse/HelloDataverse.tsx
MellyGray: Definitely, I didn't like the Trans syntax neither but among all the [alternatives](https://react.i18next.com/latest/trans-component) for the `<code>` tag I liked that one the most

Mainly because if I don't know the text behind `{t('title')}` then I don't want to know the text behind `i18nKey="description"` neither.

The alternative looks like this where the text can be seen in plain English:

```
<Trans t={t} i18nKey="description">
      Edit <code>src/sections/hello-dataverse/HelloDataverse.tsx</code> and save to reload.
</Trans>
```

Maybe it looks more beautiful to you this way, I'm open to change this

README.md
```diff
@@ -1 +1,45 @@
...
+### `npm run cy:run`
+
+Launches the Cypress test runner for the end-to-end tests. \
+If you prefer to see the tests executing in cypress you can run `npm run cy:open`
+
+### `npm run lint`
+
+Launches the linter. To automatically fix the errors run `npm run lint:fix`
+
+### `npm run format`
```
GPortas, line 39: I have tested this command and some files in the repository have been formatted. It looks like you didn't run it for the last commit. Could you run it and commit the changes?

Perhaps it would be nice to run this command on a pre-commit hook.

GPortas (changes requested):
Looks good to me! I have only left a minor comment.

JR-1991:
Looks great and everything worked well on my system 😊

One thing I wanted to ask you if instead of going purely React Router, wouldn't it make sense to choose [Next.js](https://nextjs.org)? Especially the SSR and CSR capabilities are quite nice since I guess a big portion of Dataverse will be server-side. Also the (forced) directory structure could help overview all pages. What do you think?

Other than that, great stuff! Thanks for bringing Cypress on my radar, its insane 😍

README.md
```diff
@@ -1 +1,45 @@
...
+You may also see any lint errors in the console.
+
+### `npm test`
+
+Launches the test runner for the unit tests in the interactive watch mode.
+
+### `npm run cy:run`
+
+Launches the Cypress test runner for the end-to-end tests. \
+If you prefer to see the tests executing in cypress you can run `npm run cy:open`
```
pdurbin, line 33: @JR-1991 told me that if I run `npx cypress open` I can run the tests in the Cypress GUI and see what the "hello Dataverse" webapp looks like (screenshot below). Should we document this command? It seems very useful! 😄 

Screen Shot 2023-02-16 at 12 03 45 PM

pdurbin (commented):
I left a comment about `npx cypress open`.

README.md
pdurbin: Can we run the formatter automatically with a post commit hook or whatever? That way, if people forget to run it, no problem! Git will run it for you.

MellyGray: Good point!

ESlint also has the rules of prettier, then I was confident that the lint workflow was checking te formatting too, but ESlint doesn't run on .json,.yml or md files. At the same time I realized that IntelliJ is not running prettier on .json and .yml files, so even with formatting configured on save those files were excluded.

I've modified the scripts to run prettier on json, yml and m files so now the Github Workflow will fail if this type of files are not formatted.

About the pre-commit hook I added it too, thank you for your recommendation! Now is running the lint, format and unit test scripts. It is super fast and now it prevents commits not passing lint or unit tests. Maybe we should leave the GitHub actions for pull requests only, now that the pre-commit is checking lint and unit tests

README.md
MellyGray: Hi Philip, I don't understand what you mean. The README says 'If you prefer to see the tests executing in cypress you can run `npm run cy:open`' which is the same as the `npx cypress open`. Then is already documented in the README

MellyGray:
> Looks great and everything worked well on my system 😊
> 
> One thing I wanted to ask you if instead of going purely React Router, wouldn't it make sense to choose [Next.js](https://nextjs.org)? Especially the SSR and CSR capabilities are quite nice since I guess a big portion of Dataverse will be server-side. Also the (forced) directory structure could help overview all pages. What do you think?
> 
> Other than that, great stuff! Thanks for bringing Cypress on my radar, its insane 😍

@JR-1991 great! 😊 Thank you for your comment. I can't answer directly the comment so I answer you here

I am not sure if I understood you correctly about using next.js for SSR. I think the goal is to have the SPA completely uncoupled from the backend, so no server side rendering for the frontend. Just an SPA connected to an API. Then I think React Router would work for this case.

And about the forced folders structure, I am not a big fan of having the folders structure imposed so I'll go with React Router approach. In general I think next.js will impose many things since it is a framework and React is a library. And at the moment I don't see clear advantages, maybe you were thinking about something in specific about next.js?

README.md
MellyGray: @pdurbin I wanted to do exactly that, running formatter and linter automatically and then commit the changes automatically. But doing the automatic commit is the hard part, you have to give .git writing permissions, so I don't see it really convenient

I also tried the [pre-commit](https://pre-commit.com) approach but the hooks are already provided and prettier hook is not the same as our prettier configuration :(

Finally I decided to use [pre-commit npm package ](https://www.npmjs.com/package/pre-commit) it doesn't commit the changes automatically but it does run the formatter and linter automatically on commit. Then you have to manually add those changes to your commit

Do you know of a better approach to automatically run and commit the formatting?

JR-1991:
> I am not sure if I understood you correctly about using next.js for SSR. I think the goal is to have the SPA completely uncoupled from the backend, so no server side rendering for the frontend. Just an SPA connected to an API. Then I think React Router would work for this case.

My bad, was mixing things up. Makes totally sense to use React Router, Next would be an overkill in that case 🙂

> And about the forced folders structure, I am not a big fan of having the folders structure imposed so I'll go with React Router approach. In general I think next.js will impose many things since it is a framework and React is a library. And at the moment I don't see clear advantages, maybe you were thinking about something in specific about next.js?

Yea that's true, Next imposes a lot of things and regarding the folder structure, I guess thats where tastes go apart 😄 One specific advantage I see in using Next is the flexibility it offers for CSR/SSR and since we are interfacing a database, being as close as possible to the server for some components, like displaying a dataset, could benefit performance-wise. But as said above, I was mixing things up and for an SPA Next is overkill.

README.md
pdurbin: Oh! I think I simply missed `npm run cy:open` because it didn't have `###` in front of it! My mistake! 😅 

Thanks for pointing it out! I'll resolve this.

README.md
pdurbin: Hmm. I don't but let me ping @poikilotherm because he worked hard on this for over at https://github.com/gdcc/xoai/pull/36

johannes-darms:
We are using [react-query](https://tanstack.com/query/v4/docs/react/overview) with [Axios](https://axios-http.com/) for API request and it made our lives a bit easier with refreshing, error handling and caching API request in the SPA. Maybe, its worth to explore/test those libraries for the dataverses SPA.

MellyGray:
> We are using [react-query](https://tanstack.com/query/v4/docs/react/overview) with [Axios](https://axios-http.com/) for API request and it made our lives a bit easier with refreshing, error handling and caching API request in the SPA. Maybe, its worth to explore/test those libraries for the dataverses SPA.

Hi @johannes-darms thank you for your comment! React query is really useful, I added it to the [SPA MVP document](https://docs.google.com/document/d/1vOSsIGGCbgal4f_lSp1yInQifBcmo5vZ-Z9pfkVDyG4/edit#heading=h.7fuhu31nrmj7) to take it into account for API requests

README.md
MellyGray: Thank you Philip! I finally managed to add the formatting changes to the commit automatically. I just had to add the `git add .` command to the pre-commit scripts

pdurbin:
> * [Vite](https://vitejs.dev). Uses different bundler for production and development. The development bundler is faster but more unstable. This option was discarded because it may be differences between dev and production environments because of this 2 different bundlers.
> 
> * [Create React App](https://create-react-app.dev/docs/getting-started). Uses webpack which is slower but the most stable solution right now. Really easy set up with the integrated scripts. In the future React may change the bundler from webpack to another better solution but our code doesn't know anything about the bundler so it won't be affected. The code is not coupled to the bundler.

Can we please talk a little more about Webpack vs. Vite? I am by no means an expert in this matter but mostly from podcasts I listen to, I get the impression that Vite is faster and more modern than Webpack. (The same observation about speed is already above.) I also get the impression that the "cool kids" are moving away from Webpack.

I understand the part about Create React App might switch away from Webpack to something better in the future. I don't see a direct dependency on Webpack in our package.json so it does seem like the code is not hard coded to use Webpack.

Vite was discarded because "may be differences between dev and production environments because of this 2 different bundlers"? I'm having trouble understanding what this means. In short, I'm wondering if we can use Vite. And if not, why not. 😅 

We can talk this out in Slack or in our weekly meetings, of course, I just wanted to put this here in case we're close to merging this pull request (I have no idea).

GPortas (approved):

MellyGray:
> > * [Vite](https://vitejs.dev). Uses different bundler for production and development. The development bundler is faster but more unstable. This option was discarded because it may be differences between dev and production environments because of this 2 different bundlers.
> > * [Create React App](https://create-react-app.dev/docs/getting-started). Uses webpack which is slower but the most stable solution right now. Really easy set up with the integrated scripts. In the future React may change the bundler from webpack to another better solution but our code doesn't know anything about the bundler so it won't be affected. The code is not coupled to the bundler.
> 
> Can we please talk a little more about Webpack vs. Vite? I am by no means an expert in this matter but mostly from podcasts I listen to, I get the impression that Vite is faster and more modern than Webpack. (The same observation about speed is already above.) I also get the impression that the "cool kids" are moving away from Webpack.
> 
> I understand the part about Create React App might switch away from Webpack to something better in the future. I don't see a direct dependency on Webpack in our package.json so it does seem like the code is not hard coded to use Webpack.
> 
> Vite was discarded because "may be differences between dev and production environments because of this 2 different bundlers"? I'm having trouble understanding what this means. In short, I'm wondering if we can use Vite. And if not, why not. 😅
> 
> We can talk this out in Slack or in our weekly meetings, of course, I just wanted to put this here in case we're close to merging this pull request (I have no idea).

Hi @pdurbin ! I'll try to answer you here just to keep the answer here for  future reference, but we can still discuss it in the weekly meeting

I know many people is choosing Vite over CRA, mostly because of the speed of Vite during the development

What I meant by "there may be differences between dev and production environments" is because Vite pre-bundles dependencies using [esbuilt](https://esbuild.github.io) for the local development but [it doesn't use it for the production build](https://vitejs.dev/guide/why.html#why-not-bundle-with-esbuild), so there may be inconsistencies between local and production build. It doesn't give me confidence that they do not use their esbuilt super fast solution for the real production build.

For me there was a decision, speed or stability, and I chose stability because I don't see that major difference in speed. I have used both Vite and CRA in different projects and I still haven't seen that speed difference. Stability and reliability I think is more important when speaking about a large and complex project like this one.

Moreover, CRA is designed specifically for React and it has a higher level of customisation than Vite, because Vite is designed to be a simple, fast solution, not just for React but for any project using JavaScript. I prefer that higher level of customisation because this is going to be a huge project and we don't know what we are going to need in the future, how specific do we need to get with the project build.

I hope I explain it better this time, and of course we can move to Vite if CRA still doesn't convince you 🙂

pdurbin:
> Vite pre-bundles dependencies using [esbuilt](https://esbuild.github.io) for the local development but [it doesn't use it for the production build](https://vitejs.dev/guide/why.html#why-not-bundle-with-esbuild), so there may be inconsistencies between local and production build. It doesn't give me confidence that they do not use their esbuilt super fast solution for the real production build.

For production builds, Vite uses Rollup, which is very popular. In all cases ES Modules are used so I'm not worried about inconsistencies.

> For me there was a decision, speed or stability, and I chose stability because I don't see that major difference in speed. I have used both Vite and CRA in different projects and I still haven't seen that speed difference. Stability and reliability I think is more important when speaking about a large and complex project like this one.

Dataverse is built on stable tools like PostgreSQL, Solr, and the Jakarata EE APIs so I'm easily convinced by the stability argument. However, while CRA the technology seems stable enough, CRA the project shows some warnings signs.

Just yesterday I was listening to a [podcast](https://podrocket.logrocket.com/sveltekit) where the creator of Svelte said, "You mentioned Create React App, which is essentially a single page app framework, even the React team are now trying to get rid of. It kind of codifies a lot of bad practices."

The React team is trying to get rid of CRA? Is that right? I googled and easily found  [a long comment from only a month ago](https://github.com/reactjs/reactjs.org/pull/5487#issuecomment-1409720741) by CRA co-creator Dan Abramov who had this to say:

"As the years passed, Create React App has stagnated."

"We could suggest people to migrate from Create React App to something like Vite."

He lays out five options for the future for CRA, from a huge rewrite to smaller endeavors. He says the React team is leaning toward option 5, which even mentions possibly moving to Vite under the hood:

"Option 5: Turn Create React App into a launcher

We could keep Create React App as a command, but turn it into a launcher. It would suggest a list of recommended frameworks, followed by the "classic" framework-less approach being the last option. That last "classic" approach would produce a client-only app like CRA does now (to avoid breaking tutorials), but could eventually move to Vite under the hood."

This leave me wondering, with so much talk of Vite from the creators of CRA, why not just use Vite?

Finally, to go back to the speed option, Dataverse is a big app and I do think we should be concerned about speed. Speed should be a feature. (That's what the creators of Stack Overflow said.)

> Moreover, CRA is designed specifically for React and it has a higher level of customisation than Vite, because Vite is designed to be a simple, fast solution, not just for React but for any project using JavaScript. I prefer that higher level of customisation because this is going to be a huge project and we don't know what we are going to need in the future, how specific do we need to get with the project build.

I heard you on this concern. I'm sure we'll want customization. I just don't know if using Vite will be a problem in this regard or not. I guess we'll learn more if we try Vite.

pdurbin (approved):
As discussed at standup just now, I'm very happy with this pull request overall. Great job, @MellyGray! My one outstanding concern is about using Create React App rather than Vite and (as discussed), I created a separate issue for that:

- #10

Meanwhile, I think this is ready for QA and merging. I ran the code early on but that was a couple weeks ago.

Issue #7, opened 2023-02-24
Layout component


MellyGray:
## Overview of the Feature Request

To start developing some pages for the SPA MVP we are going to need a common Layout component for all the pages. This Layout component is a wrapper with the common elements to all the pages, such as the header and the footer. All the pages will extend this component as a children.

## What kind of user is the feature intended for?
Dataverse frontend developers

## What inspired the request?
- Current work on SPA MVP Strategy
- MVP Strategy document: [link](https://docs.google.com/document/d/1vOSsIGGCbgal4f_lSp1yInQifBcmo5vZ-Z9pfkVDyG4)

## What existing behavior do you want changed?
The example page HelloDataverse.tsx will be encapsulated in the Layout component to serve as a real example.

## Any brand new behavior do you want to add to Dataverse Frontend?

-  Layout component wrapper for the future Dataverse pages.
-  Header component
-  Footer component
-  Add those components to the Storybook

## Any related open or closed issues to this feature request?

Storybook need to be set up https://github.com/IQSS/dataverse-frontend/issues/3

mreekie:
Grooming:
- Added NIH deliverable label.

PR #24, opened 2023-03-15
7 - Layout component

MellyGray:
This PR adds the Layout component to the SPA. This PR has has 2 major reasons to exist

1. The Layout is a complete component that allows us to check that the SPA Set Up is complete and it's funcional. Also, this component opens the discussion about folders structure and organisation of the code at an early stage of the MVP 
2. The Layout component has value on itself because it defines the structure for all the Dataverse pages, it defines the Header, the Footer and the body

- Closes https://github.com/IQSS/dataverse-frontend/issues/7

### Changes
- Layout.tsx component added
- Header.tsx component now is not an example for the Storybook but the real Dataverse Header
- Footer.tsx added
- Navbar.tsx added
- Bootstrap configured to work correctly with SASS
- Fixed an issue with the i18next that was responding a 404 if the browser language did not exist
- Added route.enum.ts as the enum to contain all the SPA routes
- Grid system added with the Container.tsx, Row.tsx and Col.tsx components
- Replaced the `\` by the two spaces `  ` in the README to improve the markdown

### Accessibility test error with the links
image

The links styles from the Dataverse violates some accessibility rules, we need to decide whether we:

1. Fix the link styles adding the underline and changing the color of the surrounding text and the link
2. Deactivate these accessibility rules if we think they are not necessary

This PR doesn't add any logic, it is just replicating the Dataverse styles form the Layout so I would suggest to test it in the following way:

1. Run `npm run storybook` and go to http://localhost:6006

2. Go to the Layout in the Storybook and press the Open canvas in new tab button
image

3. Go to Dataverse demo or Dataverse local server and compare if everything looks and behaves as expected regarding the layout (header, footer and main container)

4. Run the unit tests for the Layout with `npm run test`

The mockup is the original Dataverse Layout

image

Header and Footer added to the SPA

## Additional documentation:
- [React Bootstrap](https://react-bootstrap.netlify.app/getting-started/introduction/)


rtreacy:
Looks good, but we should change the link style to pass the accessibility test

rtreacy (approved):
no accessibility violations, vite, bootstrap with SASS, nabber, footer

kcondon:
It all looks good, thanks!
Found two minor things that may not be of concern:
1. The privacy policy link in the footer is incorrect. It is https://best-practices.dataverse.org/harvard-policies/harvard-privacy-policy.html but should be https://dataverse.org/best-practices/harvard-dataverse-privacy-policy
2. The link checker reported Checking formatting...
[warn] dist/assets/index-973ee37e.css
[warn] Code style issues found in the above file. Forgot to run Prettier?

kcondon:
@MellyGray Thanks! Merging now.

pdurbin:
> or we can make a call to the API

Yes, this should come from the API. The setting is called :ApplicationPrivacyPolicyUrl and is documented here: https://guides.dataverse.org/en/5.13/installation/config.html#applicationprivacypolicyurl

PR #41, opened 2023-03-29
32 - Add buttons and dropdowns of the design system

MellyGray:
## What this PR does / why we need it:
This PR adds the buttons and dropdown elements of the Dataverse design system as React components having them documented in the Storybook. This will make it easier for developers to use and maintain consistent buttons and dropdowns across the application.

## Which issue(s) this PR closes:

- Closes #32 

## Special notes for your reviewer:

I've used Bootstrap [Button](https://react-bootstrap.netlify.app/components/dropdowns/#rb-docs-content) and [Dropdown](https://react-bootstrap.netlify.app/components/dropdowns/#rb-docs-content) encapsulated in our own component.

I've made some customisation to the Bootstrap CSS styles to match the current styles of the Dataverse JSF frontend. This customisation made me add a css property with the `!important`, which is a bad practice, to override a css style added by Bootstrap js.

I believe this wouldn't be necessary if we used Bootstrap as it is, just changing the theme variables, instead of making a huge customisation. I've assumed we wanted this customisation so the new frontend looks as the one in JSF as specified in the [current Style Guide](https://guides.dataverse.org/en/latest/style/index.html) but let me know if we can stick to Bootstrap styles so I can do the changes to remove the property with the !important

## Suggestions on how to test this:

`npm run storybook`

Go to http://localhost:6006

In the UI section you can see the Button and DropdownButton components with their different states

## Does this PR introduce a user interface change? If mockups are available, please link/include them here:
This PR adds the button and dropdown of the Dataverse Design System as specified in the [Analysis doc](https://docs.google.com/document/d/1OxCQO7B-yJaMWRl8EzuflcSmIaCmVSs19aoD83PjPmA/edit#heading=h.32ocvoxblw2)

## Is there a release notes update needed for this change?:
Button and dropdown added to the Design System

## Additional documentation:

- [Design System Analysis](https://docs.google.com/document/d/1OxCQO7B-yJaMWRl8EzuflcSmIaCmVSs19aoD83PjPmA/edit#heading=h.32ocvoxblw2)
- [Current Dataverse Style Guide](https://guides.dataverse.org/en/latest/style/index.html)

PR #44, opened 2023-03-30
32 bootstrap not customized

MellyGray:

Remove Bootstrap customisation to simplify the development of new features and to not be forced to add important! rules to CSS properties

Bootstrap button with custom classes overriding Bootstrap styles to look exactly as the JSF button:
image

Bootstrap button with SASS variables overrided:
image

The problem:
Captura de pantalla 2023-03-30 a las 14 16 02

The thing is, since the JSF button has a different size than the Bootstrap button, the transformation property on the dropdown doesn't work well, so I need to edit the transformation to correctly show the dropdown menu.

This transformation property is added by Bootstrap js with the onlcick event, so is added as an inline styles, since inline styles have the greatest priority in CSS, no matter where I add a class to override this property. So I need to add an !important which is a bad practice.

Proposed solution:
Don't change the Bootstrap button size. Then no need to override styles that are added by Bootstrap javascript

In general, I think it would be easier for the development to don't override bootstrap classes with CSS classes on top, but rather overriding only the Bootstrap SASS variables. The problem with this is that the SPA frontend would look slightly different than the JSF frontend


MellyGray:
ok, done! I removed the tertiary button and the navbar will look slightly different but now the code base it's simpler, we can always add css if once we start developing pages we think it doesn't look good. But fore the moment simpler is better

Should we merge this branch and the https://github.com/IQSS/dataverse-frontend/pull/42 branch into https://github.com/IQSS/dataverse-frontend/pull/41 ? @pdurbin

pdurbin:
@MellyGray yes, please!

PR #91, opened 2023-05-09
Feature/83/create datasummary with markdown

### LINKED ISSUE #83: [Spike - Frontend] Create the DatasetSummary of the Dataset page

Implement the UI for #60.

Developers

## What inspired the request?

- [UI feature development process](https://docs.google.com/document/d/1vOSsIGGCbgal4f_lSp1yInQifBcmo5vZ-Z9pfkVDyG4/edit#heading=h.e6y69fx5m61r)

N/A

N/A

- #60.

ekraffmiller:
Creates a DataSummary component for the Dataset Page

- Closes #83 

I created a new branch off of my feature branch, to switch from SanitizedHtml to markdown (in order to save the SanitizedHtml code).  I hope that doesn't make reviewing the PR harder - let me know if there is a better manage the branching.

npm run storybook - view DataSummary component

no
no


src/sections/ui/markdown/MarkdownComponent.tsx
```diff
@@ -0,0 +1,27 @@
...
+      if (containerRef.current) {
+        ReactDOM.unmountComponentAtNode(containerRef.current)
+      }
+    }
+  }, [markdown])
+
+  return <div ref={containerRef} />
+}
+
+export default MarkdownComponent
```
MellyGray, line 27: To keep the consistency across the application we should upgrade this syntax to react 18. Some functions are deprecated. Try with:

```suggestion
import ReactMarkdown from 'react-markdown'
import { useRef } from 'react'

interface Props {
  markdown: string
}

export function MarkdownComponent({ markdown }: Props) {
  const containerRef = useRef<HTMLDivElement>(null)

  return (
    <div ref={containerRef}>
      <ReactMarkdown>{markdown}</ReactMarkdown>
    </div>
  )
}
```

src/sections/ui/sanitized-html/SanitizedHtml.tsx
```diff
@@ -0,0 +1,28 @@
...
+  html: string
+  options?: Config
+}
+
+const defaultOptions: Config = {
+  ALLOWED_TAGS: ['b', 'i', 'em', 'strong', 'a'],
+  ALLOWED_ATTR: ['href']
+}
+
+export function SanitizedHTML({ html, options }: SanitizedHTMLProps): JSX.Element {
```
MellyGray, line 14: We are not using this component anymore. We can remove this code, I mean, this work is already saved in the commits. Or if you prefer you can do what Guillermo suggested and branch off and keep this in a different branch in case we want to reuse this at some point.

src/dataset/domain/models/Dataset.ts
```diff
@@ -1,5 +1,10 @@
+import { DatasetField } from './DatasetField'
+import { License } from './License'
+
 export interface Dataset {
   id: string
   title: string
   version: string
+  summaryFields: DatasetField[]
+  license: License
 }
```
MellyGray, line 10: I suggest adding all the Dataset model interfaces to the same file. I like seeing DatasetFields and License as part of Dataset. Also since there is no logic in this kind of files, it doesn't bother anyone to see a long long file. This is just a data structure, I think it makes sense to see the whole structure at once. What do you think?

```

export interface License {
   name: string
   shortDescription: string
   uri: string
   iconUrl?: string
 }

export interface DatasetField {
   title: string
   description: string
   value: string
 }

export interface Dataset {
   id: string
   title: string
   version: string
   summaryFields: DatasetField[]
   license: License
 }

```

MellyGray (commented):
I love seeing we are making progresses with the Dataset page!

I left some comments but overall this looks good to me

src/dataset/domain/models/Dataset.ts
ekraffmiller: Thanks for the comments, Melina!  Those suggestions sound good to me, just have a question or two, will respond above 👍

src/sections/ui/sanitized-html/SanitizedHtml.tsx
ekraffmiller: I agree, I will remove this since it is saved in a previous commit - I forgot to remove it when I made the new branch

src/sections/dataset/datasetSummary/DatasetSummary.tsx
```diff
@@ -0,0 +1,47 @@
...
+          <img
+            alt={dataset.license.name + ' license icon'}
+            src={dataset.license.iconUrl}
+            title={dataset.license.shortDescription}></img>{' '}
+          <a href={dataset.license.uri}>{dataset.license.name}</a>
+        </Col>
+      </Row>
+    </article>
+  ) : null
+}
```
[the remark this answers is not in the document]
MellyGray: Good point! I don't think is necessary. We can split them in the future if we rehuse them outside of DatasetSummary

src/sections/dataset/Dataset.tsx
```diff
@@ -25,7 +26,7 @@ export function Dataset({ datasetRepository, id }: DatasetProps) {
           <Col sm={9}>Citation Block</Col>
         </Row>
         <Row>
-          <Col sm={9}>Summary Block</Col>
+          <DatasetSummary datasetRepository={datasetRepository} id={id}></DatasetSummary>
```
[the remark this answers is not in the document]
ekraffmiller: Ok, I will remove datasetRepository here, but since the License inside the DatasetSummary needs dataset data, I think I need to pass dataset, instead of dataset.summaryFields, does this look ok?

`<DatasetSummary dataset={dataset} />`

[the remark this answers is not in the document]
ekraffmiller: Or I could pass the two domain objects separately: 

`<DatasetSummary fields={dataset.summaryFields}, license={dataset.license} />`

src/sections/dataset/datasetSummary/DatasetSummary.tsx
```diff
@@ -0,0 +1,47 @@
...
+/*
+<img src="https://licensebuttons.net/p/zero/1.0/88x31.png" title="Creative Commons CC0 1.0 Universal Public Domain Dedication. " style="display: inline;" onload="this.style.display='inline'">
+ */
```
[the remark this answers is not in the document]
ekraffmiller: yes, removed 👍

pdurbin (commented):
Overall this is looking good. I hope this is a stepping stone to someday supporting Markdown!

I left a few comments.

mreekie:
sprint kickoff
- dataset summary section issue - Ellen is working on.
- size: 3 minimal work left to get it through review and QA. 
- Merge conflicts. Otherwise it's close

src/sections/dataset/Dataset.tsx
```diff
@@ -37,7 +38,9 @@ export function Dataset({ datasetRepository, id }: DatasetProps) {
               <Col sm={9}>Citation Block</Col>
             </Row>
             <Row>
-              <Col sm={9}>Summary Block</Col>
```
MellyGray, line 40: Can we add the `<Col sm={9}>` wrapping the DatasetSummary component to make the page look more like the jsf page?

ekraffmiller: yes, done 👍

src/sections/dataset/dataset-summary/DatasetSummary.tsx
```diff
@@ -0,0 +1,17 @@
...
+}
+
+export function DatasetSummary({ summaryFields, license }: DatasetSummaryProps) {
+  return summaryFields && license ? (
+    <article>
+      <SummaryFields summaryFields={summaryFields}></SummaryFields>
+      <License license={license}></License>
+    </article>
+  ) : null
+}
```
MellyGray, line 17: It seems like this props cannot be null so we don't need to check if they exists `{ summaryFields, license }`. We could get rid of `summaryFields && license ?`

I was looking at the Dataset form and the summary fields seems to be required fields so there's always going to be summaryFields. We don't need to add a check for that

The license object seems to be added automatically to any new Datasets so it cannot be null neither.

If there is a situation where they can be null we can add the check and change the props definition, otherwise we don't need to check them

The same for the License and SummaryFields components, if we have a check in this first level we don't need a check in those components

ekraffmiller: ok, thanks, I will remove the check for null, and remove the undefined type from the props

jggautier: Sorry for the noise if I'm misunderstanding, but are you saying that the metadata fields that appear near the top of the dataset page below the blue citation box and above the tabs will never be null or empty?

<kbd>Screenshot 2023-05-16 at 8 02 06 AM</kbd>

In that screenshot, keywords appear, but Harvard Dataverse doesn't force depositors to add keywords. And the [customdatasetsummaryfields](https://guides.dataverse.org/en/5.13/installation/config.html?highlight=customdatasetsummaryfields#customdatasetsummaryfields) setting let's any Dataverse installation decide which metadata fields appear in that section.

ekraffmiller: @MellyGray I removed the null check from License and SummaryFields, but when I tried to remove it from DataSummary, the component test failed, saying the map variable within SummaryFields was undefined. 
So I committed the change with the null check removed for License & SummaryFields, let me know what you think

ekraffmiller: I just looked at the code https://github.com/IQSS/dataverse/blob/fc23042ab337dafb1c92f2aab15c73a223bedcec/src/main/java/edu/harvard/iq/dataverse/dataset/DatasetUtil.java#L44 and if custom summary fields is null or empty, it returns a default list of fields, so I think we don't have to worry about that scenario.

MellyGray: > @MellyGray I removed the null check from License and SummaryFields, but when I tried to remove it from DataSummary, the component test failed, saying the map variable within SummaryFields was undefined.
> So I committed the change with the null check removed for License & SummaryFields, let me know what you think

I was looking at the code and the error about the` the map variable within SummaryFields was undefined` is because the DatasetMother mock needs to be updated to contain the summaryFields and license. I just did a commit to add those fields to DatasetMother and remove the check from DatasetSummary.tsx

I'm going to approve this, thanks!

MellyGray (approved):
Looks good to me! Approved! 🎉 

Reminder: we need to first merge the Boilerplate

kcondon:
@ekraffmiller I didn't see the DataSummary component in story book. Probably I'm overlooking something obvious. I did see a "summary block" label in the sample dataset page.

Issue #63, opened 2023-04-20
Add all downloading options to Dataset page in view mode


GPortas:
## Overview of the Feature Request

The goal is to add all downloading options to Dataset page in view mode.

This include: citation, files, and metadata (See mockups).

### Mockup of the requested element

![metadata_export](https://user-images.githubusercontent.com/17010447/233394740-233168cc-764c-4b86-b37b-68d9aa44d6cc.png)
![citation_block](https://user-images.githubusercontent.com/17010447/233394745-ca583b7c-a577-4d31-8f02-ada8889c83f8.png)
![file](https://user-images.githubusercontent.com/17010447/233394755-71e8a93a-f219-46a4-b6fe-0faf00d7304d.png)

### Process action items

_Find out more about the action items in the [process definition](https://docs.google.com/document/d/1vOSsIGGCbgal4f_lSp1yInQifBcmo5vZ-Z9pfkVDyG4/edit#heading=h.e6y69fx5m61r)_

- [ ] Analyze required data access logic
- [ ] Document associated API endpoints
- [ ] Develop the UI element
- [ ] Add the element to the page and storybook
- [ ] Create spike issues if required and add them to _related open or closed issues_

## What inspired the request?

- UI discussion on Dataset page initial work

## What existing behavior do you want changed?

N/A

## Any brand new behavior do you want to add to Dataverse?

- Add Dataset downloading options to Dataset page

## Any open or closed issues related to this feature request?

Not yet

pdurbin:
<img width="343" alt="Screen Shot 2023-04-25 at 11 35 43 AM" src="https://user-images.githubusercontent.com/21006/234328958-7b5ff945-801d-4ff2-b932-0e3fea59d7a7.png">

Please note that for the "Cite Dataset" dropdown we need to create an API endpoint for the various formats (Endnote XML, RIS, and BibTeX):

- https://github.com/IQSS/dataverse/issues/3140

I'm not sure why I gave up on this, but long ago (2016!) I did start hacking a bit on adding this citation API: https://github.com/IQSS/dataverse/commit/c433395f4a1756ee14e669b91820047edc4db1ad

PR #148, opened 2023-07-18
136 - Files table UI [2/2] - Row selection 

### LINKED ISSUE #136: [Spike - Frontend] [2/2] Create the File Tab of the Dataset - Checkboxes selection + selection message

Add the checkboxes selection + the selection message at the top of the Files table

image

Developers

## What inspired the request?

- #105

None

Add the checkboxes selection + the selection message at the top of the Files table UI

- #105

MellyGray:
This PR adds the row selection feature to the UI of the files table. Since the library we're using already has a row selection feature, in this PR I only had to add a few checks to show 2 types of messages related with the row selection. The message about the selection and the message about the zip download limit exceeded. 

This is only the UI no backend connected yet

In this other issue we'll connect the row selection with the download button:

- #63 

- Closes #136 

Since the library we're using already has a row selection feature, in this PR I only had to add 2 types of messages related with the row selection. The message about the selection and the message about the zip download limit exceeded

Also this is the first time we're using settings so I created a SettingsContext to get the backend settings from the API from any react component

This is only the UI no backend connected yet

1. `npm i`
2. `cd packages/design-system && npm run build`
3. cd back to the root 
4. `npm run storybook`
5. Check the files table at localhost:6006 storybook in the Dataset page
6. Test the row selection going through the pages selecting different rows
7. Select a few rows and check that a message appears informing about the selection with 2 buttons to select All or to clear selection, ensure this 2 buttons work
8. Select a few rows with a total file size greater than 500 bytes (mocked value) to see the message about the zip download limit

To check the steps in video you can go to https://www.loom.com/share/e970a9abbe6841b59c2e18937532319e?sid=6db11224-ef41-4988-938b-5cc8ce0a57d2

Done!

image

Row selection added to the Files Table from the Dataset page


GPortas:
I have tested Storybook and row selection works as expected.

We need to decide whether to keep the JSF behavior, which only shows the "limit exceeded" message if more than one file has been selected, but doesn't show it when only one has been selected, even if just that single file has exceeded the limit.

One file:

files_select_1_file

Two files:

files_select_2_file

src/sections/dataset/dataset-files/files-table/row-selection/RowSelectionMessage.tsx
```diff
@@ -0,0 +1,51 @@
...
+import styles from './RowSelectionMessage.module.scss'
+
+interface RowSelectionMessageProps {
+  selectedFilesCount: number
+  totalFilesCount: number
+  setRowSelection: (rowSelection: RowSelection) => void
+}
+
+const MINIMUM_SELECTED_FILES_TO_SHOW_MESSAGE = 0
+const MINIMUM_FILES_TO_SHOW_MESSAGE = 10
```
GPortas, line 13: Just a note, no changes required. In the future it may be interesting to convert these constants to config properties of the application.

src/settings/domain/models/Setting.ts
```diff
@@ -0,0 +1,8 @@
+export enum SettingName {
+  ZIP_DOWNLOAD_LIMIT = 'ZIP_DOWNLOAD_LIMIT'
```
GPortas, line 2: We should remove this SettingName enum once it is included in the js-dataverse package. The package will export it just like for the use cases. So we don't duplicate it in the SPA side.

src/settings/domain/models/ZipDownloadLimit.ts
```diff
@@ -0,0 +1,3 @@
+import { FileSize } from '../../../files/domain/models/File'
+
+export class ZipDownloadLimit extends FileSize {}
```
GPortas, line 3: Do you really need this extra `ZipDownloadLimit` class?  I assume you include it for better code readability.

GPortas (commented):
I have left some comments. Looks good overall!

GPortas:
@MellyGray There is a wrong copy in one of the filters.

SPA ("Filter Type"):
spa_filter

JSF ("File Type"):
jsf_filter

It should be "File Type".

src/settings/domain/models/Setting.ts
MellyGray: I always duplicate the types in the SPA, this way my imports look like
`import { Setting, SettingName } from '../settings/domain/models/Setting'`

instead of:
import { Setting,SettingName } from '@iqss/dataverse-client-javascript'`

Using the domain definitions instead of types from an external library so the code is not coupled to an external library. 

Just imagine that I need to add a new setting to the UI, if I were using js-dataverse type definition I'd would have to wait for this type to be updated in js-dataverse in order to use it in the SPA. By having my own SettingsName I can add new settings without waiting for js-dataverse to be updated, I mean, this is how I can develop new UI features without waiting for the backend/js-dataverse to be ready

src/settings/domain/models/ZipDownloadLimit.ts
MellyGray: Yes, I like being able to declare my use case as `getSettingByName<ZipDownloadLimit> `instead of `getSettingByName<FileSize> `. Yep, better readability, I think it changes the useCase meaning completely

src/sections/dataset/dataset-files/files-table/row-selection/RowSelectionMessage.tsx
MellyGray: If we want to give that level of customisation to the files table, yes, we could make all of these part of a config file 👍

MellyGray:
@GPortas I answered to your comments and made the changes about the Filter Type typo, glad that you notice that, thanks 👍 

note: the e2e GitHub action is failing but it's not related to the code changes

GPortas (approved):

Issue #127, opened 2023-06-27
Add all alert messages related to the Dataset View Mode


MellyGray:
## Overview of the Feature Request
We need to add all the different alert messages related to the Dataset View mode

These are some examples of the Alert messages

![image](https://github.com/IQSS/dataverse-frontend/assets/23359572/8a80bb82-ac0a-459a-b561-10323e5a1134)

## What kind of user is the feature intended for?

Default user role since we are not handling any user role at the moment

## What inspired the request?
We are implementing the Dataset page and we need to show different alert messages for each state. 

## What existing behavior do you want changed?
None

## Any brand new behavior do you want to add to Dataverse?
Show alert messages in the Dataset view depending on each dataset states

## Any open or closed issues related to this feature request?
- #99

PR #190, opened 2023-10-03
Feature/99 alert mechanism

### LINKED ISSUE #99: Add alert messages management mechanism to the UI
We need an alert message mechanism in the UI to show the different alert messages to the user.

These are some examples of the Alert messages

image

Default user role since we are not handling any user role at the moment

## What inspired the request?
We are implementing the Dataset page and we need to show different alert messages for each state. 

None

Alert messages management

- #58

ekraffmiller:
This PR adds logic to the JSDatasetMapper that will create DatasetAlert objects to be used in the Dataset View page
 
- Closes #99 

Run the Cypress JSDatasetMapper unit test


tests/component/sections/dataset/dataset-files/files-table/files-info/file-info-cell/file-info-data/FileDate.spec.tsx
```diff
@@ -1,10 +1,16 @@
...
+    const fileDate = new Date('2023-09-18')
+    const date = { type: FileDateType.PUBLISHED, date: fileDate }
     cy.customMount(<FileDate date={date} />)
-
-    cy.findByText(`Published Sep 18, 2023`).should('exist')
+    const dateString = fileDate.toLocaleDateString(Intl.DateTimeFormat().resolvedOptions().locale, {
+      year: 'numeric',
+      month: 'short',
+      day: 'numeric'
+    })
```
MellyGray, line 13: Since this dateString is depending on the user's locale, this and the other changed tests are failing for me, because the FileDate component is implemented to always show the US date format and the test is checking the user's locale format, which in my case is the Spanish format

So either we change the test to check for the US date format or we refactor the FileDate component to show the user's date format.

In JSF I think it shows always the US format so that's why I implemented this like that, but I think that changing it to the user's locale could be an improvement, should we discuss this with Jullian?

src/dataset/infrastructure/mappers/JSDatasetMapper.ts
```diff
@@ -29,19 +35,22 @@ export class JSDatasetMapper {
...
+      ),
+      privateUrl
     ).build()
   }
 
   static toVersion(
     jDatasetVersionId: number,
-    jsDatasetVersionInfo: JSDatasetVersionInfo
+    jsDatasetVersionInfo: JSDatasetVersionInfo,
+    requestedVersion?: string
```
MellyGray, line 46: There are some e2e tests that are checking the DatasetVersion that need to be fixed because of the added requestedVersion property

MellyGray (commented):
I really like the implementation of the Dataset Alerts as part of the Dataset model 🎉 I just left a few minor comments

I was also thinking about the other alerts, the ones that are not related with the Dataset Model but with certain behaviours of the user, such as 'there was an error when you clicked some button' Are we planning to implement those in the [ other alerts issue](https://github.com/IQSS/dataverse-frontend/issues/127) or should we create a new issue for this type of alerts not related with domain models?

tests/component/sections/dataset/dataset-files/files-table/files-info/file-info-cell/file-info-data/FileDate.spec.tsx
ekraffmiller: Ok, I see what you mean about locale format, that makes sense.  I think it's a good idea to use the user's  locale format.  I made the change because my local time zone was showing a different day since I'm five hours ahead, but that is fixed by testing against a dynamic date string, rather than changing the locale.
I can put a message in dv-tech to confirm about using the user's locale format, since it's different from the JSF

ekraffmiller:
Hi Melina, I think I addressed all your change requests - the e2e test is failing, but in a different way on the github action than in my local environment - maybe that's because they are including different dataverse images?

MellyGray:
> Hi Melina, I think I addressed all your change requests - the e2e test is failing, but in a different way on the github action than in my local environment - maybe that's because they are including different dataverse images?

@ekraffmiller the FileEmbargoDate component tests are still failing for me
image

Regarding the e2e tests running in the Github action, I had the same question and Guillermo told me that the Github Action is pointing to the dataverse develop branch image because ideally the backend changes should be merged before we start working on the frontend. But that's not the case right now, so some tests are failing in the GitHub action. 

But apparently it's ok, now the backend PR's have been given priority, so hopefully the backend will be ahead of the frontend soon

So I ran the e2e locally and I'm having this error, but I already fixed it in another branch:

image

Then some e2e tests fail from time to time because of the wait(), Cypress doesn't like being waited for so long, sometimes it crashes, we'll have to address that at some point but for this PR, I think it's ok like this.

In summary, can you take a look at the unit tests for the FileEmbargoDate component? the e2e test are ok like this

ekraffmiller:
Hi Melina, Gustavo suggested I create a spike for the question of the dates being returned by the API, so I created this issue https://github.com/IQSS/dataverse/issues/9979.  I think the locale logic is separate from the timezone issue, and we have agreement on using the user's locale, so I don't think this PR is dependent on that, but let me know what you think.

MellyGray (approved):
LGTM 😄

PR #227, opened 2023-11-13
224 - Integration files tabular data

MellyGray:
## What this PR does / why we need it:
This PR integrates the files tabular information from the js-dataverse module into the SPA repository to display the files tabular data in the Files Table.

## Which issue(s) this PR closes:

- Closes #224 

## Special notes for your reviewer:
A part from the tabular data I tried to integrate the files `deleted` property but this property is not working in the js-dataverse module as expected since it always returns undefined

### Step 1: Run the development environment
1. Run `npm I`
3. `cd packages/design-system && npm run build`
4. `cd ../../`
5. Check that you have a  `.env` file such as the .env.example, with the `VITE_DATAVERSE_BACKEND_URL=http://localhost:8000` variable
6.  `cd dev-env`
7. `./run-env.sh unstable`
8. To check the environment go to <http://localhost:8000> and check your local dataverse installation

### Step 2: Test Dataset View mode with the implemented changes for the getFilesTotalDownloadSize
1. Go to <http://localhost:8000>
2. Login as admin using username: dataverseAdmin and password: admin1
3. Create a new Dataset
4. Upload a tabular file to the dataset such as a .csv
5. From the dataset view mode copy the search parameters from the url. Ex.: `?persistentId=doi:10.5072/FK2/LHGRHP&version=DRAFT`
6. Go to <http://localhost:8000/spa/datasets> and paste the search parameters. Ex.: <http://localhost:8000/spa/datasets?persistentId=doi:10.5072/FK2/LHGRHP&version=DRAFT>
8. You are seeing now the same Dataset in the SPA. Check that the tabular data of the file is being displayed as expected: variables, observations, unf...

## Does this PR introduce a user interface change? If mockups are available, please link/include them here:
No

## Is there a release notes update needed for this change?:
No

PR #240, opened 2023-11-28
238 - Integration single file download

### LINKED ISSUE #238: Single file download
downloadfile

There is already an endpoint under the access API to download a single file. This endpoint is consumed from JSF. See: https://github.com/IQSS/dataverse/blob/develop/src/main/java/edu/harvard/iq/dataverse/FileDownloadHelper.java#L88

We can follow the same approach as for image thumbnails, without using js-dataverse and using direct API call.

SPA users.

## What inspired the request?
- https://github.com/IQSS/dataverse-frontend/issues/191#issuecomment-1759380984

None

Single file download

- https://github.com/IQSS/dataverse-frontend/issues/191#issuecomment-1759380984

MellyGray:
This pull request establishes a connection between the file download buttons in the files table rows and the access API for downloading files.

- Closes #238 

The integration of files tabular data is currently under review:
- https://github.com/IQSS/dataverse-frontend/pull/227

As a result, the tabular files' download options won't appear identical to those in the JSF frontend until the aforementioned pull request is merged. Ideally, PR #227 should be merged into the develop branch before this one so that the tabular download options align with the JSF frontend.

1. Run `npm I`
4. `cd ../../`
6.  `cd dev-env`
7. `./run-env.sh unstable`

### Step 2: Test Dataset View mode with the implemented changes to download the files
3. Create a new Dataset
4. Upload some files to the dataset
5. From the dataset view mode copy the search parameters from the url. Ex.: `?persistentId=doi:10.5072/FK2/LHGRHP&version=DRAFT`
6. Go to <http://localhost:8000/spa/datasets> and paste the search parameters. Ex.: <http://localhost:8000/spa/datasets?persistentId=doi:10.5072/FK2/LHGRHP&version=DRAFT>
8. You are currently viewing the same dataset in the SPA. In the files table, locate the "Access File" dropdown on the right of each row. From the dropdown, select the option to download the file and verify that it is successfully downloaded.

No

No


GPortas:
@MellyGray Please, can you resolve conflicts?

MellyGray:
@GPortas conflicts resolved

GPortas (approved):
Looks good!

downloadsinglefile

GPortas:
https://github.com/IQSS/dataverse-frontend/assets/17010447/1c780e11-c4d2-4741-9493-1194000dca72

MellyGray:
> @MellyGray
> 
> Tabular files work well for both original and archival formats. But when selecting RData the following error appears:
> 
> {"status":"ERROR","code":404,"message":"datafile access error: requested optional service (image scaling, format conversion, etc.) could not be performed on this datafile."}

@GPortas The same error appears in JSF when trying to download a tabular file in RData format. 

I found this in the [documentation](https://guides.dataverse.org/en/latest/installation/prerequisites.html?highlight=rdata):

> The Dataverse Software uses [R](https://cran.r-project.org/) to handle tabular data files. The instructions below describe a minimal R Project installation. It will allow you to ingest R (.RData) files as tabular data and to export tabular data as .RData files. **R can be considered an optional component**, meaning that if you don’t have R installed, you will still be able to run and use the Dataverse Software - but the functionality specific to tabular data mentioned above will not be available to your users.

I think we need to install R in the dev-env if we want to test that feature, although I don't think it's necessary since if the users install that option, it should work as expected

GPortas:
@MellyGray In my opinion the download option in R format should not be rendered if R is not configured in Dataverse. To do this we could use a public setting that allows the SPA to know if R is configured in the Dataverse instance.

It certainly happens that when testing JSF on my localhost, I see that the R option is available, although clicking it does nothing. I think this behavior should not be expected for the user. Even though JSF handles it this way, I think the SPA should use the approach I described above (conditional rendering based on a public setting).

Maybe @scolapasta, @jggautier or @pdurbin can share their thoughts on this and if there is a reason that explains why it is done that way in JSF.

GPortas (changes requested):
By the way, I think the conditional rendering solution I described above should not block the PR at all. Whatever solution we propose, I would implement it in another issue.

For this PR I think we should catch the error that occurs (black screen) when clicking the R option, so that it does nothing (like in JSF) but doesn't break. @MellyGray

MellyGray:
> By the way, I think the conditional rendering solution I described above should not block the PR at all. Whatever solution we propose, I would implement it in another issue.
> 
> For this PR I think we should catch the error that occurs (black screen) when clicking the R option, so that it does nothing (like in JSF) but doesn't break. @MellyGray

@GPortas  I can't handle the error since it's an href. Besides, the same black screen behavior is happening in JSF, so I think it should be okay like this until we decide to implement hiding the RData option if R is not installed.

https://github.com/IQSS/dataverse-frontend/assets/23359572/96f15b4d-9fda-4b01-bc4e-e0f7855da60d

GPortas (approved):

pdurbin:
> In my opinion the download option in R format should not be rendered if R is not configured in Dataverse.

Yes, @BPeuch and probably others agree with you. That's almost exactly how he worded it:

- https://github.com/IQSS/dataverse/issues/8711

PR #241, opened 2023-11-29
236 cypress timing error

### LINKED ISSUE #236: Cypress timing error in Dataset e2e test
There is a timing error that happens intermittently in the e2e test, so far I've only seen it happening in Github, not locally.  It is in Dataset.spec.tsx, in the 'loads the restricted files when the user is logged in as owner', test.  Attached is the log information from the gihub action. The Cypress error message suggests that there is an action chained after click() in the code, but the test doesn't have that, so there must be other code triggering it.

Here is an action where it is happening: https://github.com/IQSS/dataverse-frontend/actions/runs/6905396784.
Attached are details from the Github action log.
Screenshot 2023-11-20 at 1 59 28 PM
Screenshot 2023-11-20 at 1 59 07 PM

ekraffmiller:
The e2e Dataset test was sometimes failing, due to asynchronous page updates that happened while the test was running

- Closes #236 

Because of the timing nature of this error, I couldn't reproduce it consistently, so I can't be sure that this change fixes the problem. But I did break down the steps in the test into smaller increments, based on the Cypress recommendations for this type of error. I've run the test a couple of times, and it seems to be working. Note that I've only seen this happen in the Github action, not when running the test locally.
Make sure that the e2e test in the github Action passes.  Run it a couple of times to try to trigger the error

no
no
## Additional documentation:
https://docs.cypress.io/guides/references/error-messages#cy-failed-because-the-page-updated

### DIFF
--- a/src/files/domain/models/File.ts
+++ b/src/files/domain/models/File.ts
@@ -1,4 +1,5 @@
 import FileTypeToFriendlyTypeMap from './FileTypeToFriendlyTypeMap'
+import { FileUserPermissions } from './FileUserPermissions'
 
 export enum FileSizeUnit {
   BYTES = 'B',
@@ -115,6 +116,7 @@ export enum FileLabelType {
   CATEGORY = 'category',
   TAG = 'tag'
 }
+
 export interface FileLabel {
   type: FileLabelType
   value: string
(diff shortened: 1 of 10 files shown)


ekraffmiller:
This last commit shows that the error was still happening, even after moving the API call to earlier in the page loading logic.

I thought at first the the double rendering and Effects caused by React StrictMode were causing the error, but I was able to reproduce the error with StrictMode removed.  During my testing I was able to consistently reproduce the error, and fixed the error by breaking up the Cypress commands before clicking the button.  

Given that the refactoring didn't work, I'm going to close this PR and create a new one that just makes the changes to the Cypress test.

PR #255, opened 2023-12-11
236 - Cypress timing error

### LINKED ISSUE #236: Cypress timing error in Dataset e2e test

ekraffmiller:
In the e2e test, the Dataset tests are sometimes failing because the page is asynchronously updated during the Cypress call to click the AccessFileMenu.  To avoid this, the Cypress command for finding the button and the call to click the button are separated into two statements.
here is Cypress documentation on the issue: https://docs.cypress.io/guides/references/error-messages#cy-failed-because-the-page-updated

- Closes #236 

Due to the timing nature of the error, it's hard to reproduce consistently.  I was able to get my local dev environment in a state where it was happening by adding some console.log statements, and when I applied the fix, the error no longer happened.  
Run the e2e test locally and check the e2e test on Github Actions.  

no
no

### DIFF
--- a/tests/e2e-integration/e2e/sections/dataset/Dataset.spec.tsx
+++ b/tests/e2e-integration/e2e/sections/dataset/Dataset.spec.tsx
@@ -277,11 +277,14 @@ describe('Dataset', () => {
 
           cy.findByText('Restricted File Icon').should('not.exist')
           cy.findByText('Restricted with access Icon').should('exist')
-
-          cy.findByRole('button', { name: 'Access File' }).should('exist').click()
+          cy.findByRole('button', { name: 'Access File' }).as('accessButton')
+          cy.get('@accessButton').should('exist')
+          cy.get('@accessButton').click()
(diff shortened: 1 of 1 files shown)


MellyGray (approved):
I re-run the e2e tests several times both on my machine and on the Github actions and they seem to be passing every time so I'm going to approve

Issue #26, opened 2023-03-17
[Docs] Dataverse Frontend Guide


GPortas:
## Overview of the Feature Request
Goal: Set up the documentation guides for the Dataverse Frontend project, similar to the general sphinx-based Dataverse guides, but specific to the Frontend application.

- Documentation will be hosted on Git as a README or series of README files as needed.

A part from the general project documentation, there are some specific TODO's (to document) from PR discussions that need to be addressed:

- [x]  Explain Object Mother pattern in the documentation and link Martin Fowler article as discussed in this PR:   https://github.com/IQSS/dataverse-frontend/pull/260#discussion_r1451605405

## What kind of user is the feature intended for?
- Dataverse users and frontend developers

## What inspired the request?
- Recent needs to document frontend procedures. For example: application deployment and accessibility changes. We are currently documenting these procedures in the project README, though they will eventually be migrated to the guides once they are established.
- Discussion with @pdurbin 
- #16

## What existing behavior do you want changed?
N/A

## Any brand new behavior do you want to add to Dataverse Frontend?
N/A

## Any related open or closed issues to this feature request?
- #16

pdurbin:
> We don't necessarily have to use sphinx for these guides. Let's do research on possible implementations to find the best fit.

There's active discussion about Sphinx vs. other options for a relaunched version of https://dataversecommunity.global (which is currently an OpenScholar/Drupal site but might become a static site at https://github.com/gdcc/dataversecommunity.global ):

https://dataverse.zulipchat.com/#narrow/stream/375707-community/topic/migrate.20GDCC.20github.20pages

To summarize the tools mentioned so far:

- Sphinx
- Sphinx with Markdown using MyST: https://myst-parser.readthedocs.io
- Jekyll
- mkdocs: https://squidfunk.github.io/mkdocs-material/getting-started/
- Docusaurus (see https://github.com/IQSS/dataverse/issues/8905 )

pdurbin:
In tech hours yesterday I briefly presented the "Sphinx with Markdown using MyST" option in the context of fixing https://dataversecommunity.global by switching it from OpenScholar (where everything but the homepage is [broken](https://dataversecommunity.global/about)) to something, anything on GitHub Pages. Here's the code so far: https://github.com/gdcc/dataversecommunity.global/tree/sphinx-markdown

I thought @ekraffmiller made a good point: is the README enough for now? It makes me question if this should be in "sprint ready" or not. 🤔 

[...]

PR #260, opened 2023-12-13
249 - File Page boilerplate

### LINKED ISSUE #249: Set up the boilerplate for the File Page
Set up the boilerplate for the FilePage.

Features:

1. File page with the File title and a reference to it's dataset
2. The files from in the Dataset page files table should have a link to it's file page
3. Loading page
4. Page not found if the file is not found

## Mock
Captura de pantalla 2023-12-05 a las 12 01 26
image
image

All users

## What inspired the request?

None

New File Page in the SPA beta

MellyGray:
This PR adds the boilerplate of the File Page. A basic page to display the files details. The boilerplate includes these features:

Features:

1. File page with the File title and a reference to it's dataset
2. The files in the Dataset page files table should have a link to it's file page
3. Loading page
4. Page not found if the file is not found

- Closes #249 

This is just the most basic version of the page. We'll add more components to the page in other issues.

I had to refactor the Dataset mode so the DatasetVersion property could be extracted and added to the File model since all of the dataset version data will be needed to implemented the File page

8. Visit [the File Page in the Storybook](http://localhost:6006/?path=/story/pages-file--default)

## Mock
Captura de pantalla 2023-12-05 a las 12 01 26
image
image

File Page added to the SPA beta


pdurbin:
I took a quick peek at https://646f68aa9beb01b35c599acd-naljpjbrwa.chromatic.com/?path=/story/pages-file--default

Screenshot 2023-12-13 at 3 07 34 PM

MellyGray:
I'm putting this back to 'in progress' because I need to refactor the Dataset model to include the dataset version in the new File model. We'll need some dataset version information to display it on the file page. This refactor will impact all the upcoming file page issues, so I think it's better to do the refactor in this branch, which will serve as the parent for the rest of the file components branches

src/files/domain/models/File.ts
```diff
@@ -1,207 +1,6 @@
...
-    if (this.embargo) {
-      return this.embargo.isActive
-    }
-    return false
-  }
-
-  get isTabularData(): boolean {
-    return this.tabularData !== undefined
-  }
+export interface File {
```
GPortas, line 3: It is a bit confusing that the class with the largest number of properties is the "preview", while "File" is only an interface with two properties.

GPortas (changes requested):
Overall it looks good

I left a comment about a naming issue.

On the other hand, storybook looks good for Default and Loading states:

file_default

file_loading

But for File Not Found, I see a reference to "Demo Dataverse" which is not correct. I assume it was introduced by mistake due to a string copied from the demo installation.

src/files/domain/models/File.ts
MellyGray: I see your point, but the File model is not finished yet, and when the File Page is finished the File model will contain all the FilePreview properties plus the file versions, plus the file metrics, etc. So at the end the File model will be larger than the File Preview, even though right now it's smaller because we only have the boilerplate properties.

Also, the way I see it is that in the Files Table of the Dataset View we're displaying the file information preview, while in the File View we're seeing the actual file.

Does it make sense to you?

MellyGray:
> But for File Not Found, I see a reference to "Demo Dataverse" which is not correct. I assume it was introduced by mistake due to a string copied from the demo installation.
> 

I removed the last sentence of the alert to match the message in beta:

image

image

src/files/domain/models/File.ts
GPortas: Makes sense. Thanks!

GPortas (approved):
LGTM

src/stories/dataset/dataset-action-buttons/DatasetActionButtons.stories.tsx
```diff
@@ -27,15 +27,14 @@ export const WithPublishPermissions: Story = {
     <DatasetActionButtons
       dataset={DatasetMother.create({
         permissions: DatasetPermissionsMother.createWithAllAllowed(),
-        version: DatasetVersionMother.createDraftAsLatestVersion(),
+        version: DatasetVersionMother.createDraftAsLatestVersionWithSomeVersionHasBeenReleased(),
```
M27Mangan, line 30: Can you walk me through what this method does? I would like to try for a shorter name that is still as descriptive.

M27Mangan (changes requested):

src/stories/dataset/dataset-action-buttons/DatasetActionButtons.stories.tsx
MellyGray: This method creates a mock of the dataset version, the name tries to describe that this dataset version is a draft and it's the latest version available of the dataset + some version of the dataset has been released

This means that the dataset is published and the retrieved version is a draft which is also the latest version available

tests/component/files/domain/models/FilePreviewMother.ts
```diff
@@ -0,0 +1,384 @@
...
+      ingest: FileIngestMother.createInProgress()
+    })
+  }
+
+  static createIngestProblem(reportMessage?: string): FilePreview {
+    return this.createDefault({
+      ingest: FileIngestMother.createIngestProblem(reportMessage)
+    })
+  }
+}
```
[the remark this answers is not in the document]
MellyGray: The 'Mother' portion of the name simply refers to the [Object Mother pattern](https://java-design-patterns.com/patterns/object-mother/). It's not precisely a Factory or a Builder class, which is why it carries its unique pattern designation. 

I understand your concerns regarding the gender implications of the name. However, including other names would alter its meaning.

We could include an explanation about this pattern and its naming in the documentation.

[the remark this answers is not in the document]
pdurbin: I also asked about "mother" and I do think it would be nice to link to that Martin Fowler article from a comment or other docs. Good idea.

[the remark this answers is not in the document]
MellyGray: @M27Mangan correct me if I'm wrong but I think you were working on this issue:

- https://github.com/IQSS/dataverse-frontend/issues/26

Should we add the Martin Fowler article to that guide or should I add it to the README?

The problem with linking the article from a comment is that there are several files where the Mother suffix is used so I'm not sure where to place it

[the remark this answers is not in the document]
pdurbin: Maybe just the README for now? Or in a list of TODOs (to document) in this issue?

- #26

GPortas:
I just retested the branch and it looks good, merging.

PR #269, opened 2023-12-21
261 - Add File Citation to File Page

### LINKED ISSUE #261: Add the FileCitation section to the File Page
Add the FileCitation section to the File Page to display the citation information of the file and the dataset. 

Features:
1. Add FileCitation component to the File page
2. Add DatasetCitation component to the File page

## Mock
image

All users

## What inspired the request?

None

None

- #260

MellyGray:
This PR adds the File Citation section to the UI of the File Page.

- Closes #261 

I refactored some components related to the citation to rehuse them in the FileCitation component

The citation boxes won't look as in the mock, with the coloured background because in the SPA the dataset citation box only has the border coloured, so I did the same with the File Citation box.

8. Visit [the File Page in the Storybook](http://localhost:6006/?path=/story/pages-file--default)
9. For detailed stories of the FileCitation component you can go to the [FileCitation section of the storybook](http://localhost:6006/?path=/story/sections-file-page-filecitation--default) 

image

No


tests/component/files/domain/models/FileMother.ts
```diff
@@ -4,18 +4,35 @@ import { DatasetVersionMother } from '../../../dataset/domain/models/DatasetMoth
...
+      name: 'File Title',
       datasetVersion: DatasetVersionMother.createRealistic(),
+      citation: FileCitationMother.create('File Title'),
       ...props
     })
   }
 }
+
+export class FileCitationMother {
+  static create(fileName: string): string {
```
GPortas, line 27: I think here is where the string is composed incorrectly. (${fileName} [fileName]`)

src/sections/dataset/dataset-citation/DatasetCitation.tsx
```diff
@@ -1,56 +1,41 @@
...
+            isDeaccessioned={version.publishingStatus === DatasetPublishingStatus.DEACCESSIONED}
+          />
+        </Col>
+      )}
+      <Col sm={withoutThumbnail ? 12 : 10}>
+        <CitationDescription
+          citation={version.citation}
+          tooltip={<DatasetCitationTooltip status={version.publishingStatus} />}
+        />
+        <CitationLearnAbout />
```
GPortas, line 37: Nice extraction

GPortas (changes requested):
In general it looks good.

I have found a detail that I believe is an error in the composition of the file citation block text.

filecitation_err

I don't see anything similar in JSF

filecitationjsf

jggautier:
@GPortas, about the composition of the file citation block text, could you see what the citation looks like when the file has a persistent ID?

All files in Demo Dataverse, including the file in your "JSF" screenshot, are given a persistent ID.

But for files that aren't given persistent IDs, such as https://dataverse.harvard.edu/file.xhtml?fileId=8079268&version=1.0, it looks like the file's citation block text composition is the same as your first screenshot with the red circle.

GPortas:
Thank you @jggautier,

Then we have to compose the citation text by choosing one of the two composing options depending on whether the file has a persistent ID or not.

@MellyGray I think we should add this distinction in the logic in this PR.

tests/component/files/domain/models/FileMother.ts
MellyGray: This is how I see the files in JSF, with the [fileName] part. If you look at the [DataCitation.java](https://github.com/IQSS/dataverse/blob/871f3735a1344fd78a506c6acdb7aa84a55f5d8c/src/main/java/edu/harvard/iq/dataverse/DataCitation.java#L218C87-L218C87) in the dataverse repo, that part is added if there is no global identifier

image


In any case, I was assuming that the citation was coming from the API, as in the Dataset page. So this `(${fileName} [fileName]`)` logic only applies to the Stoybook, then I can change it to a different string if we don't like the one with the [fileName]

GPortas:
@MellyGray is right. The citation string must come from the js-dataverse package, as for datasets, so the way it is composed must be opaque to the SPA. This is something to address in the upcoming API and js-dataverse issues and not here.

GPortas (approved):

ekraffmiller:
@MellyGray can you merge the conflicts in this PR?

ekraffmiller:
looks good!

PR #277, opened 2024-01-09
253 - Fix do not show pagination if there is only 1 page + files results copy

### LINKED ISSUE #253: Pager should not be shown when there are not enough items + wrong message for files count

Pager should not be shown when there are not enough items.

On the other hand, the SPA should replicate JSF for the files count title: "1 File" instead of "1 to 1 of 1 Files". We should double check how the copy looks like for more file counts to make sure the behavior is the same.

I encountered this while testing: https://github.com/IQSS/dataverse-frontend/pull/227

SPA:
spatab

JSF:
jsftab

All of them

## What inspired the request?
- https://github.com/IQSS/dataverse-frontend/pull/227

Pager should not be shown when there are not enough items

None

- https://github.com/IQSS/dataverse-frontend/pull/227

MellyGray:
This PR fixes 2 bugs in the Files Table:

1 - Pagination controls shouldn't be shown if the total number of files would result in just  1 page. The expected behaviour is showing the pagination controls when there is more than 1 page of files.
2 - When there is only 1 File in the table the results copy should be ✅  '1 File' instead of ❌ '1 to 1 of 1 Files'

- Closes #253 

1. Run `npm I`
4. `cd ../../`
6.  `cd dev-env`
7. `./run-env.sh unstable`

### Step 2: Test Dataset View mode with the implemented changes to download the files
3. Create a new Dataset
4. Upload some files to the dataset
5. From the dataset view mode copy the search parameters from the url. Ex.: `?persistentId=doi:10.5072/FK2/LHGRHP&version=DRAFT`
6. Go to <http://localhost:8000/spa/datasets> and paste the search parameters. Ex.: <http://localhost:8000/spa/datasets?persistentId=doi:10.5072/FK2/LHGRHP&version=DRAFT>
8. You are currently viewing the same dataset in the SPA. In the files table, check the copy and the pagination bugs. Upload or remove files as necessary to test the bug.

No

No


tests/component/sections/dataset/dataset-files/DatasetFiles.spec.tsx
```diff
@@ -116,7 +116,7 @@ describe('DatasetFiles', () => {
       cy.findByRole('columnheader', { name: '101 to 150 of 200 Files' }).should('exist')
     })
 
-    it('renders the first page if there is only one page and the user changes to a lower page size', () => {
+    it('does not render pagination if the user changes to a lower page size resulting in one page', () => {
```
M27Mangan, line 119: Shouold this be "... if the user changes to a **_larger_** page size..." ?

M27Mangan (changes requested):
I would like to also explore hiding the buttons for "First" and "Last" if there are only two pages.

MellyGray:
> I would like to also explore hiding the buttons for "First" and "Last" if there are only two pages.

@M27Mangan Why would you want to hide those? They exist in the JSF frontend. Furthermore, it could be confusing for users to see the buttons change just for two pages. While the "First" and "Last" buttons may be redundant with only two pages, I believe it's better to maintain this redundancy rather than introduce too many exceptions to the User Interface, which can be confusing for users

image

tests/component/sections/dataset/dataset-files/DatasetFiles.spec.tsx
MellyGray: Thanks, I fixed the text

M27Mangan:
Review complete, moving to Ready for QA!

ekraffmiller:
looks good! Merging

PR #281, opened 2024-01-16
263 - Add file metadata section to the File page

### LINKED ISSUE #263: Add the FileMetadata section to the File Page
Add the FileMetadata section to the File Page to display the file metadata information 

Features:
1. Add the FileMetadata component

## Mock
Captura de pantalla 2023-12-14 a las 11 38 09

All users

## What inspired the request?

None

None

- #260

MellyGray:
This PR adds the file metadata section UI of the File Page.

- Closes #263 

8. Visit [the File Page in the Storybook](http://localhost:6006/?path=/story/pages-file--default) and check that the File Metadata section looks as expected

File Metadata section added to the File Page
Captura de pantalla 2023-12-14 a las 11 38 09

File Metadata section added to the File Page


MellyGray:
During the development I realised that it would be great to extract the file metadata to its own model so it can be used via composition in both the File and the FilePreview model so I opened a separate PR to review the refactor https://github.com/IQSS/dataverse-frontend/pull/281

src/sections/file/File.module.scss
M27Mangan, line 1: Instead of adding style references in scss, have you considered using the spacing utilities in the [Bootstrap Documentation](https://getbootstrap.com/docs/5.3/utilities/spacing/) so then as an alternative to adding to the `.row` global style, we could add the classnames `my-1` and `py-3` for `.row` and `.tab-container` respectively? 
Just a thought, I know it's late in the game.

src/sections/file/file-labels/FileLabels.module.scss
M27Mangan, line 1: If this scss file only contains one file, we could similarly add a `<div className={mr-2}>` as an inner container as another opportunity to use the existing Boostrap spacing utils referenced in above comment.

src/sections/file/File.module.scss
MellyGray: Good question! Just for clarification, the `.row` you're seeing in this .scss file is not the global class `.row` from Bootstrap; it's a custom style that I created specifically for the File component. When rendered, it becomes class .row_b2323212, so there's no overriding of the Bootstrap `.row` class, and no conflict with any other `row` class from other .scss files, since all are rendered differently.

Regarding the suggestion of using `py-3`,`.row` and `.tab-container` classes directly , let me explain the rationale behind our current approach. There was a discussion about this at the beginning of the project.

Here are the advantages of using styles like `styles.row`, `styles.subtext`, etc., over directly using Bootstrap classes in the components:

1. These are not global styles. When the component is rendered, the `.row` class from the .scss file becomes something like `.row_b2323212`, ensuring that the classes are only applied where intended. This eliminates the possibility of a .row class being unexpectedly applied to a distant div in the component, without your knowledge.
2. Using `styles.row` enhances component readability. You might have encountered components like `<div className="tab-container grid-1 left mb-7 blablabla bbalabalb ablabaalabla" />`, where 90% of the code are classes that add nothing to the component's functionality and make the component logic hard to read.
3. It ensures separation of concerns: the component code focuses on component logic, and the .scss file on CSS. So, for example, when I want to give styles to the subtext component, I encapsulate all the styles in `styles.subtext` because, when you're reading the component logic, you don't need to know about whether I'm applying some color, margin, or padding to this part of the component. If you're interested, you can refer to the .scss file.
4. We can still use some Bootstrap classes, such as row and container, but only in their component form. So, instead of using the row bootstrap class, we use the `<Row></Row>` component, or the `<Container></Container>` component if you want to use the container class

I hope that answers you question. I believe this is in some part of the initial docs of the SPA but I'm not sure where

src/sections/file/file-labels/FileLabels.module.scss
MellyGray: I addressed this in the above comment, but to summarize: we prefer to encapsulate all the styles for separation of concerns and to enhance readability. `styles.container` encapsulates whatever style is being applied to the container of the FileLabels. This is because when I'm reading the component logic, I'm not concerned with what classes are being applied, which improves readability.

Also, if in the future we want to add another style property, we won't need to modify the component file, just the .scss file. This is as it should be, since we're only modifying styles. In a way, we're following the open/closed principle

M27Mangan (approved):
Everything seems to be good on my end, and thank you again for taking the time to answer my questions!

ekraffmiller:
looks good! Merging

PR #284, opened 2024-01-18
280 - Use page search param in home page pagination

### LINKED ISSUE #280: Use Query Param in Home Page dataset list pagination
From Guillermo' QA comments, steps to reproduce:

**TEST 4 - Pagination: Back button and URL page query param**

🚨

In JSF, a query parameter is added to the URL to indicate the results page you are on. I think we should continue to support this. It is useful for scenarios like the following:

- When you have navigated to a new result page, by clicking on the browser's back button you navigate to the previously selected page.

- With the query param you are also able to share a URL that leads directly to a particular page on the home page.

You can see in the following video how, by not adding the query parameter, the back button navigates to the page before reaching the home page.

https://github.com/IQSS/dataverse-frontend/assets/17010447/618710aa-a48d-4472-9e57-8cd76ab49c71

_Originally posted by @GPortas in https://github.com/IQSS/dataverse-frontend/pull/274#pullrequestreview-1813474292_

MellyGray:
This PR adds the functionality of navigating the different pages of the datasets list in the home using the url search params.

- Closes #280 

1. Run `npm I`
4. `cd ../../`
6.  `cd dev-env`
7. `./run-env.sh unstable`

### Step 2: Test the Home
3. Create as many datasets are needed to have pagination in the home page
4. Navigate through the pages using the buttons and check that the page search param corresponds to the current page
5. Open a new tab and paste `http://localhost:8000/spa?page=2` to access directly the page 2 using the url

No

No

### DIFF
--- a/src/sections/home/Home.tsx
+++ b/src/sections/home/Home.tsx
@@ -4,9 +4,10 @@ import { DatasetRepository } from '../../dataset/domain/repositories/DatasetRepo
 import { DatasetsList } from './datasets-list/DatasetsList'
 interface HomeProps {
   datasetRepository: DatasetRepository
+  page?: number
 }
 
-export function Home({ datasetRepository }: HomeProps) {
+export function Home({ datasetRepository, page }: HomeProps) {
   const { t } = useTranslation('home')
 
   return (
@@ -14,7 +15,7 @@ export function Home({ datasetRepository }: HomeProps) {
       <header>
         <h1>{t('title')}</h1>
       </header>
-      <DatasetsList datasetRepository={datasetRepository} />
+      <DatasetsList datasetRepository={datasetRepository} page={page} />
     </Row>
   )
 }
--- a/tests/e2e-integration/e2e/sections/home/Home.spec.ts
+++ b/tests/e2e-integration/e2e/sections/home/Home.spec.ts
@@ -6,39 +6,59 @@ describe('Home Page', () => {
   const title = faker.lorem.sentence()
   before(() => {
     TestsUtils.setup()
+    TestsUtils.login()
   })
 
-  beforeEach(() => {})
+  beforeEach(() => {
+    TestsUtils.login()
+  })
+
   it('successfully loads', () => {
     cy.visit('/spa')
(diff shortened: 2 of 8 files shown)


src/sections/home/Home.tsx
ekraffmiller, line 1: The back button doesn't appear to be working as expected, see example here, where the page results stay the same.  Is that expected behavior?
https://github.com/IQSS/dataverse-frontend/assets/675224/1181fe37-cc3f-4a05-9b96-ae343727f32d

tests/e2e-integration/e2e/sections/home/Home.spec.ts
```diff
@@ -6,39 +6,79 @@ describe('Home Page', () => {
...
-    TestsUtils.login()
-    cy.findByText(/Dataverse Admin/i).should('exist')
-    cy.findByText(title).should('be.visible')
-    cy.findByText(title).click({ force: true })
-    cy.url().should('include', 'persistentId')
-    cy.findAllByText(title).should('be.visible')
+  it('navigates to a dataset from the list when clicking the title', () => {
+    cy.wrap(
+      DatasetHelper.destroyAll().then(() => DatasetHelper.createWithTitle(title)),
+      { timeout: 6000 }
```
ekraffmiller, line 24:
```suggestion
      { timeout: 10000 }
```
The test was failing because the timeout expired before destroyAll() completed.  We have another ticket for adding destroyAll() to more tests, to avoid having too many to delete at once, but as quick fix, increasing the timeout to 1000 worked for me locally

ekraffmiller (changes requested):
Overall looks good, I just have some questions about the behavior.

src/sections/home/Home.tsx
MellyGray: Thanks for your suggestions!

I added a fix so that the back and forward buttons from the browser can be used to navigate through the pages.

Also, I added a Page Number Not Found error to the page to be displayed when the page does not exist

ekraffmiller (approved):
The changes look good! Approved

GPortas (approved):
Works good! Merging.

PR #290, opened 2024-01-26
272 move filepreview permissions

### LINKED ISSUE #272: Move loading of File Permissions out of FilePermissionsProvider
There are still some timing errors happening when the File Permissions are retrieved in the FilePermissionsProvider, so we will move this logic to the main file loading function in FileJSDaverseRepository.ts.  This will also be consistent with how dataset permissions are loaded in the View Dataset use case.

ekraffmiller:
The original purpose of this PR was to the file permissions hook, which was believed to be causing a timing error in Cypress e2e tests.
After refactoring the code to remove the hook, the error is still occuring! For example in this run:
https://github.com/IQSS/dataverse-frontend/actions/runs/7671731412/job/20910744807

I did some more reading about Cypress handling of click actions, and found this article, which explains that sometimes click() fails because Cypress clicks the button before the event handler is attached:
https://www.cypress.io/blog/2019/01/22/when-can-the-test-click.

This article suggests using cypress-pipe to retry the click until the event handler is attached, but I couldn't get it to work in Electron, and when I tried to switch headless mode to Chrome I ran into another issue: https://github.com/cypress-io/cypress/issues/14175

So I added a cy.wait() after finding the button, and before clicking, and that fixed it.
I added another library, cypress-grep, that allows a flaky test to be run multiple times.  Without the cy.wait(), it would fail about 25% of the time, and with cy.wait() it succeeded for up to 50 runs, so I think that fixed it.

Also on this PR another unrelated Cypress failure is happening, I created an issue for it here https://github.com/IQSS/dataverse-frontend/issues/291

- Closes #272


package.json
```diff
@@ -107,6 +107,7 @@
     "@auto-it/released": "10.46.0",
     "@csstools/postcss-cascade-layers": "3.0.1",
     "@cypress/code-coverage": "3.10.7",
+    "@cypress/grep": "^4.0.1",
```
MellyGray, line 110: Can you remove the ^ symbol to avoid unexpected upgrades?

src/files/infrastructure/FileJSDataverseRepository.ts
```diff
@@ -49,15 +50,17 @@ export class FileJSDataverseRepository implements FileRepository {
           jsFiles,
           FileJSDataverseRepository.getAllDownloadCount(jsFiles),
           FileJSDataverseRepository.getAllThumbnails(jsFiles),
+          FileJSDataverseRepository.getAllWithPermissions(jsFiles),
```
MellyGray, line 53: If the file permissions are already obtained here, then we can remove the `getUserPermissionsById(id: number)` method from this file and from the FileRepository declaration

Also I think we can remove the checkFileDownloadPermission and checkFileEditDatasetPermission use cases

src/files/domain/models/FilePreview.ts
```diff
@@ -1,4 +1,5 @@
 import FileTypeToFriendlyTypeMap from './FileTypeToFriendlyTypeMap'
+import { FilePermissions } from './File'
```
MellyGray, line 2: I noticed that the `FilePermissions` model is actually duplicated. There is already a `FileUserPermissions` model, we could delete the current `FilePermissions` from the File.ts file and then rename `FileUserPermissions`  to `FilePermissions` and then use it everywhere

tests/component/sections/dataset/dataset-files/files-table/file-actions/file-actions-cell/file-action-buttons/access-file-menu/AccessFileMenu.spec.tsx
```diff
@@ -52,7 +41,7 @@ describe('AccessFileMenu', () => {
     cy.findByText('Public').should('exist')
   })
 
-  it('renders the request access button', () => {
+  it.only('renders the request access button', () => {
```
MellyGray, line 44:
```suggestion
  it('renders the request access button', () => {
```

tests/e2e-integration/integration/files/FileJSDataverseRepository.spec.ts
```diff
@@ -90,6 +91,7 @@ describe('File JSDataverse Repository', () => {
         .getAllByDatasetPersistentId(dataset.persistentId, dataset.version)
         .then((files) => {
           files.forEach((file, index) => {
+            console.log('algorithm', file.checksum?.algorithm)
```
MellyGray, line 94:
```suggestion

```

tests/e2e-integration/integration/files/FileJSDataverseRepository.spec.ts
```diff
@@ -99,6 +101,8 @@ describe('File JSDataverse Repository', () => {
             cy.compareDate(file.date.date, expectedFile.date.date)
             expect(file.downloadCount).to.deep.equal(expectedFile.downloadCount)
             expect(file.labels).to.deep.equal(expectedFile.labels)
+            console.log('algorithm', file.checksum?.algorithm)
+            console.log('expectedFile', expectedFile.checksum?.algorithm)
```
MellyGray, line 105:
```suggestion

```

tests/support/e2e.ts
```diff
@@ -20,3 +20,9 @@ import { DataverseApiAuthMechanism } from '@iqss/dataverse-client-javascript/dis
 
 const VITE_DATAVERSE_BACKEND_URL = (import.meta.env.VITE_DATAVERSE_BACKEND_URL as string) ?? ''
 ApiConfig.init(`${VITE_DATAVERSE_BACKEND_URL}/api/v1`, DataverseApiAuthMechanism.SESSION_COOKIE)
+
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore
+import registerCypressGrep from '@cypress/grep'
+// eslint-disable-next-line @typescript-eslint/no-unsafe-call
+registerCypressGrep()
```
MellyGray, line 28: I couldn't find where the grep library is being used, is it to use the cy.wait(500) command in the integration test?

MellyGray (changes requested):
Looks good, this refactor gives me such a peace of mind, I think the permissions code it's much easier to read this way

I left a few comments, the main one being that we need to merge the FilePermissions and the FileUserPermissions model

tests/support/e2e.ts
ekraffmiller: I used it from the command line when I was testing locally. For example, 
` npx cypress run --spec tests/e2e-integration/e2e/sections/dataset/Dataset.spec.tsx --browser chrome --env grep="loads the restricted files when the user is logged in as owner",burn=10`

Even though it's not a part of the automated test, I thought it would be good to have in the project, to make it easier to run tests like this locally, for timing errors.  Does that sound ok, or should I keep it part of my local environment?

src/files/domain/models/FilePreview.ts
ekraffmiller: I also removed canEditDataset from the FilePermissions interface and also the enum, because it didn't seem to be needed, let me know if that's ok.

src/stories/file/FileWithDeniedPermissionsRepository.ts
```diff
@@ -1,17 +1,17 @@
...
+import { FilePermissions } from '../../files/domain/models/FilePermissions'
 import { FileMockRepository } from './FileMockRepository'
 
 export class FileWithDeniedPermissionsRepository
   extends FileMockRepository
   implements FileRepository
 {
   // eslint-disable-next-line unused-imports/no-unused-vars
-  getUserPermissionsById(id: number): Promise<FileUserPermissions> {
+  getUserPermissionsById(id: number): Promise<FilePermissions> {
```
MellyGray, line 11: We can delete this method since we removed getUserPermissionsById from the FileRepository

src/stories/file/FileWithGrantedPermissionsRepository.ts
```diff
@@ -1,17 +1,17 @@
...
+import { FilePermissions } from '../../files/domain/models/FilePermissions'
 import { FileMockRepository } from './FileMockRepository'
 
 export class FileWithGrantedPermissionsRepository
   extends FileMockRepository
   implements FileRepository
 {
   // eslint-disable-next-line unused-imports/no-unused-vars
-  getUserPermissionsById(id: number): Promise<FileUserPermissions> {
+  getUserPermissionsById(id: number): Promise<FilePermissions> {
```
MellyGray, line 11: We can delete this method since we removed getUserPermissionsById from the FileRepository

tests/component/files/domain/models/FilePreviewMother.ts
```diff
@@ -24,7 +24,22 @@ const createFakeFileLabel = (): FileLabel => ({
   type: faker.helpers.arrayElement(Object.values(FileLabelType)),
   value: faker.lorem.word()
 })
+export class FilePermissionsMother {
```
MellyGray, line 27: FilePermissionsMother class already exists at `tests/component/files/domain/models/FilePermissionsMother.ts`

tests/support/e2e.ts
MellyGray: Sounds good! We could document this command in the [Dataverse Frontend Guide](https://github.com/IQSS/dataverse-frontend/issues/26) so that everyone is aware that tests can be run with this command

MellyGray (changes requested):
Looks good! Just a few remaining methods that need to be removed

MellyGray (approved):
LGTM!

GPortas:
@ekraffmiller 

Can you resolve merge conflicts?

If I am not wrong, they are related to https://github.com/IQSS/dataverse-frontend/pull/285 merge

PR #325, opened 2024-02-25
305 update chromatic configurations

### LINKED ISSUE #305: Update Chromatic configurations
Ensure that Chromatic tests are using consistent data provided by the Faker library. Chromatic should be re-reviewed as well

Referenced from:

> It's nice to have all the reviews up to date! I noticed that some changes are being flagged because we are using randomized test data, it might be useful to modify how the Stories are run for chromatic, so that randomized doesn't get interpreted in Chormatic as a change in the UI.
> For example in this review- 
> https://www.chromatic.com/review?appId=646f68aa9beb01b35c599acd&number=296&type=linked&view=changes
> 
> In the files table, we randomize File Access and ingest state
> Image
> 
> And in multiple stories, we have random faker data, for example
> Image
> 
> We could use a faker.seed() to have reproducible results (https://fakerjs.dev/guide/usage.html#reproducible-results) and have conditional logic to only set the seed for a chromatic run https://www.chromatic.com/docs/ischromatic/#with-an-environment-variable.
> It could be a bit involved to set the seed correctly and have it visible to all the stories, so it might be worth saving that as a separate ticket and keep doing a visual review for now.  Would be good to discuss with the team.
> 
> _Originally posted by @ekraffmiller in https://github.com/IQSS/dataverse-frontend/issues/270#issuecomment-1933249850_
>

ekraffmiller:
The Chromatic snapshots are hard to review because they show changes that are due to faker randomized data, and some stories have built in loading delays.  The changes make the faker data more reproducible, and add a delay before taking some snapshots.

- Closes #305 

To reproduce the behavior of a Chromatic build, add a` .env `file to the root directory of the project, with this setting:
`STORYBOOK_CHROMATIC_BUILD=true`
In a local build, open Storybook and go to a story.  If you refresh the browser multiple times, the faker data should stay the same between refreshes.
The link to the Chromatic build is in the github action. You can go to the build, and approve the current changes (there have been ui changes since the last approval), and check that the faker data doesn't change between builds.


M27Mangan:
@ekraffmiller - Chromatic works as-intended now, its a marvelous thing! 
Just need to fix the merge conflicts and we're good to go!

M27Mangan (changes requested):
Finishing file review and requesting changes

ekraffmiller:
Hi @M27Mangan, I merged the latest from develop, and you can see that several snapshots have changed now...I think that's because the new collection stories changed the order in which faker data was generated.  I experimented with setting the faker seed in every .stories file to avoid this, but didn't have much luck. So this is a compromise, and sometimes the data will change. I think if we want to pursue it more, we could create another issue, let me know what you think.

M27Mangan (approved):

M27Mangan:
👍  Approved!

ekraffmiller:
@MellyGray I updated the code to remove the loading delay, because I wanted to make sure all the calls to faker were made in a deterministic way. This removed most of the differences, but there are still some differences in FileInfoCell that I need to figure out.  So I'm going to put the issue back to In Progress

MellyGray (approved):
Looks good! The Chromatic builds seem to be fixed!

PR #351, opened 2024-03-21
337 multiple values field type

### LINKED ISSUE #337: Convert fields from the Create Dataset Form to multiple values field type

Certain fields in the Create Dataset Form are intended to support multiple values. This design allows them to include Add or Remove buttons, enabling users to input additional values for the same field.

The purpose of this issue is to introduce logic that transforms simple fields into multiple-value fields.

Captura de pantalla 2024-03-11 a las 10 25 21

It should be noted that we have some implementation of this multi-value field feature in our design system. However, it is not functioning as expected. Extracting the feature to make it functional and reusable across the design system appears to be challenging. Consequently, we may need to remove this multi-value field feature from the design system. Instead, we could provide a story as an example of how to implement it, although it's likely that we will not have a dedicated component for this feature due to the need for looping within the form.

This feature is intended for all users.

## What inspired the request?
- Suggestions from the beta proposal labeled as "nice to haves."

The desired change is to introduce logic that transforms simple fields into those that can accept multiple values.

No new behavior is proposed for addition to Dataverse at this time.

- #316

ekraffmiller:
Hard-coded version of multiple metadata values input, to be modified for the dynamic Create Dataset form

- Closes #337


src/sections/create-dataset/useDatasetFormData.tsx
```diff
@@ -1,22 +1,46 @@
...
+    console.log('added field' + JSON.stringify(updatedFormData.metadataBlocks[0].fields['author']))
+  }
+  const removeField = (path: string, index: number) => {
+    const updatedFormData = _.cloneDeep(formData)
+    const arrayAtPath: string[] = _.get(updatedFormData, path) as string[]
+    arrayAtPath.splice(index, 1)
+    _.set(updatedFormData, path, arrayAtPath)
+    setFormData(updatedFormData)
+    // datasetIsValid(updatedFormData)
+  }
```
MellyGray, line 43: It appears that this is no longer in use. Could you remove it? I was confused when reviewing the code, as I couldn't understand why this logic appeared in two different places

src/sections/create-dataset/AuthorFormGroup.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+import { ChangeEvent } from 'react'
+import _ from 'lodash'
+import { useMultipleFields } from './useMultipleFields'
+interface AuthorFormGroupProps {
+  submissionStatus: SubmissionStatus
+  initialAuthorFields: DatasetMetadataSubField[]
+}
+
+export function AuthorFormGroup({ submissionStatus, initialAuthorFields }: AuthorFormGroupProps) {
```
MellyGray, line 16: I believe we need to pass updateFormData as a prop and invoke it within handleFieldChange, since this function is responsible for updating the form data

ekraffmiller: thanks for your comments, @MellyGray I made the changes you suggested

src/stories/create-dataset/AuthorFormGroup.stories.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+const meta: Meta<typeof AuthorFormGroup> = {
+  title: 'Sections/Create Dataset Page/AuthorFormGroup',
+  component: AuthorFormGroup,
+  decorators: [WithI18next, WithLayout]
+}
+export default meta
+type Story = StoryObj<typeof AuthorFormGroup>
+
+export const Default: Story = {
```
MellyGray, line 17: This is how I see the field. I think `Author Name ?` should be in the same line. Also the add button is not aligned with the field.

image

MellyGray (changes requested):
Looks good 🎉 

I've just added a few suggestions for minor improvements.

Also, should we remove the dynamic field from the design system?

src/sections/create-dataset/dynamic-fields-buttons/DynamicFieldsButtons.tsx
```diff
@@ -0,0 +1,38 @@
...
+}
+
+export function DynamicFieldsButtons({
+  originalField,
+  onAddButtonClick,
+  onRemoveButtonClick
+}: AddFieldButtonsProps) {
+  return (
+    <div className={styles.container}>
+      <Tooltip placement="top" overlay="Add">
```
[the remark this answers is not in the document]
ekraffmiller: I added translation text to createDataset.json, in the future when we have an edit form, maybe that can be refactored into a separate file for both create and edit forms.

src/sections/create-dataset/AuthorFormGroup.tsx
```diff
@@ -0,0 +1,78 @@
...
+import { ChangeEvent } from 'react'
+import _ from 'lodash'
+import { useMultipleFields } from './useMultipleFields'
+interface AuthorFormGroupProps {
+  submissionStatus: SubmissionStatus
+  initialAuthorFields: DatasetMetadataSubField[]
+  updateFormData: (name: string, value: string | DatasetMetadataSubField[]) => void
+}
+
+export function AuthorFormGroup({
```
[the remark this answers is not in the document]
ekraffmiller: done!  I noticed that findByLabelText() wasn't working, I think because the "Author Name" wasn't attached to the input.  So I created another label, 'Name' that is grouped with the input field.  This also matches what we want for compound fields.

src/stories/create-dataset/AuthorFormGroup.stories.tsx
ekraffmiller: thanks, I fixed the alignment problems

ekraffmiller:
Thanks @MellyGray I made the changes you suggested. The e2e test is currently failing on get  total dataset counts, which the clear test data PR should fix.

MellyGray (approved):
LGTM!

GPortas (commented):
There is a failing e2e test. I'm not sure if this is a known error with an associated issue to fix it or if the error was introduced in this PR.

e2eeerrr

I have tested the create dataset form with the new multiple field in storybook and it looks good! But I have found some issues.

https://github.com/IQSS/dataverse-frontend/assets/17010447/44e12853-d685-4ecc-8c07-adca5b747b5d

1) After the initial load, why is it the only field that appears with a red border, if the others are also required?

2) After clicking the "add" button, to add one more author name, the other required fields show their red border.

3) I've been playing with "add" and "delete" buttons. You can see how after removing the second author name when we have three authors, the form is automatically submitted.

ekraffmiller:
Thanks @GPortas for finding that form submission bug!  I fixed that, and updated the validation to use the same hook as the other fields, so now I think the behavior is consistent with the rest of the form.
The e2e tests are failing because of race conditions when clearing test data, which I think will be fixed by this PR #360

GPortas (commented):
Thanks for the addressed changes, @ekraffmiller 

Maybe the following issues are not very relevant now because the field validation is going to change soon. Possibly not to address in this PR, but at least to keep in mind. Please let me know what you think please: 

The second author field does not show a required field message if it is left blank, anyway if we add a third author field and then remove it and leave the second blank, the required field message appears (I reproduce this in the recording below).

https://github.com/IQSS/dataverse-frontend/assets/17010447/4a7aed4d-a90e-4e25-acee-d442e0e9458f

> if you add an author field, but keep it blank, the validation doesn't flag it, but I think that's ok because it is requiring at least one author

In JSF, if you add a multiple field that is required and leave it blank, the required field message is displayed. I think we should replicate this.

jsfauthors

GPortas (approved):

Issue #294, opened 2024-01-31
Clear test datasets before integration test 


ekraffmiller:
When the integration tests cause a build up of datasets in Dataverse, then calling destroyAll() before the "gets the total dataset count" test in DatasetJSDataverseRepository.spec.ts causes an optimistic lock exception in Dataverse.

To avoid having a large number of datasets to delete all at once, we can add destroyAll() to the beforeEach method of integration tests. 
Here is an example of the test failure:
https://github.com/IQSS/dataverse-frontend/actions/runs/7653145563/job/20968488247

PR #354, opened 2024-03-26
339 add alerts for create dataset

### LINKED ISSUE #339: Add Alerts of the Create Dataset Form

Add all Alerts related to the Create Dataset Form.
Captura de pantalla 2024-03-11 a las 10 45 40
Captura de pantalla 2024-03-11 a las 10 46 49

This feature is intended for all users.

## What inspired the request?

- Suggestions from the beta proposal labeled as "nice to haves."

None

Add all Alerts related to the Create Dataset Form.

- #316

ekraffmiller:

- Closes #339

Sometimes the e2e getTotalDatasetsCount() fails, because of this issue: https://github.com/IQSS/dataverse-frontend/issues/294

## Suggestions on how to test this
Review Dataset 'Created' Story, Create a Dataset and check for Validation Alerts for missing required fields, and check for Success message on Dataset Page after dataset is created.

see mockups in #339

### DIFF
--- a/public/locales/en/createDataset.json
+++ b/public/locales/en/createDataset.json
@@ -5,6 +5,10 @@
     "content": "After adding the dataset, click the Edit Dataset button to add more metadata."
   },
   "requiredFields": "Asterisks indicate required fields",
+  "validationAlert": {
+    "title": "Validation Error",
+    "content": "Required fields were missed or there was a validation error. Please scroll down to see details."
+  },
   "datasetForm": {
     "fields": {
       "title": {
--- a/public/locales/en/dataset.json
+++ b/public/locales/en/dataset.json
@@ -82,6 +82,10 @@
       "heading": "Success!",
       "alertText": "This dataset draft has been deleted."
     },
+    "datasetCreated": {
+      "heading": "Success!",
+      "alertText": "This dataset has been created."
+    },
     "metadataUpdated": {
       "heading": "Success!",
       "alertText": "The metadata for this dataset has been updated."
--- a/src/alert/domain/models/Alert.ts
+++ b/src/alert/domain/models/Alert.ts
@@ -11,6 +11,7 @@ export enum AlertMessageKey {
   TERMS_UPDATED = 'termsUpdated',
   THUMBNAIL_UPDATED = 'thumbnailUpdated',
   DATASET_DELETED = 'datasetDeleted',
+  DATASET_CREATED = 'datasetCreated',
   PUBLISH_IN_PROGRESS = 'publishInProgress'
 }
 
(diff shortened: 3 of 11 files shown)


ekraffmiller:
@MellyGray I added some logic to the submitForm() function to save the "created" state before navigating to the View Dataset page, but I wasn't sure where to use the location hook on that page.  I tried adding it to the DatasetFactory or the DatasetProvider, but the logic seemed a little awkward there. Do you have a idea for where logic like this should go?

MellyGray:
> @MellyGray I added some logic to the submitForm() function to save the "created" state before navigating to the View Dataset page, but I wasn't sure where to use the location hook on that page. I tried adding it to the DatasetFactory or the DatasetProvider, but the logic seemed a little awkward there. Do you have a idea for where logic like this should go?

@ekraffmiller  I would place the logic to read the state from the location hook in the DatasetFactory, simply because it's where we're reading the rest of the navigation parameters. I guess what you mean by 'awkward' is because usually, the Dataset Alerts come from the repository, so we need to pass the state all the way through the repository. That's the part that seems awkward to me. Maybe we can pass the 'created' state to the Dataset.tsx component as a new prop and then use the useAlertContext to add a new Dataset Alert from the Dataset.tsx component instead of from the repository. We could call the useAlertContext in the DatasetFactory component, but then we wouldn't be able to test it individually. I don't have a strong opinion on this

ekraffmiller:
> > @MellyGray I added some logic to the submitForm() function to save the "created" state before navigating to the View Dataset page, but I wasn't sure where to use the location hook on that page. I tried adding it to the DatasetFactory or the DatasetProvider, but the logic seemed a little awkward there. Do you have a idea for where logic like this should go?
> 
> @ekraffmiller I would place the logic to read the state from the location hook in the DatasetFactory, simply because it's where we're reading the rest of the navigation parameters. I guess what you mean by 'awkward' is because usually, the Dataset Alerts come from the repository, so we need to pass the state all the way through the repository. That's the part that seems awkward to me. Maybe we can pass the 'created' state to the Dataset.tsx component as a new prop and then use the useAlertContext to add a new Dataset Alert from the Dataset.tsx component instead of from the repository. Alternatively, we could call the useAlertContext in the DatasetFactory component, but then we wouldn't be able to test it individually. I don't have a strong opinion on this

Thanks @MellyGray, I decided to add a 'created' param to Dataset.tsx, to keep the repository logic simpler

MellyGray (approved):
LGTM! 🎉

GPortas (approved):
Looks good!

https://github.com/IQSS/dataverse-frontend/assets/17010447/53d19c6a-2ce6-4b9a-8c8f-76e47dca3c8b

Issue #347, opened 2024-03-15
Add eslint-plugin-react-hooks


GermanSaracca:
We need to add this plugin to eslint to prevent us from errors or bad practices when using different react hooks.
Such as useEffect, useMemo, useCallback, useState, etc.
Since it has not been used before, we may or may not see many warnings or errors at first.
If there are many we could simply make it show a warning and fix them as we encounter these errors when doing some other feature.

[eslint-plugin-react-hook](https://www.npmjs.com/package/eslint-plugin-react-hooks)

PR #359, opened 2024-04-01
347 Add eslint plugin react hooks

GermanSaracca:
Adds Eslint plugin for React Hooks.
We need it to enforce [Rules of Hooks](https://react.dev/reference/rules/rules-of-hooks).

- Closes https://github.com/IQSS/dataverse-frontend/issues/347

The plugin raised only a few **_errors_** when initially configured, only related to calling hooks conditionally, all were resolved.
There are several minor _**warnings**_ that we could fix as we come across them.
I only hide via a comment of eslint two errors regarding the use of a hook in the Story of the DatasetAlert as I think it might take more time than expected and it is just a story, however we could see how to configure the alerts story to avoid the eslint error in the future.
- Run tests to check everything works as usual.
- Locate a useEffect call after a conditional statement in a react component, it should show an eslint error.
- Use a local state variable or a prop inside a useEffect and don't add it to the dependency array of it, it should show an eslint warning.
No
No


GermanSaracca:
Yes, it makes total sense, we could tackle those warnings as well. I'm going to resize this issue due to the large amount of warnings (23) received about incorrect dependencies in useEffect which is something I need to be careful with because otherwise we can end up in infinite loops, multiple calls to the api or unnecessary re-renders.

GermanSaracca:
As discussed with the team, we thought the best idea would be to resolve the remaining warnings in a separate issue.
At the moment we dropped from having 7 errors and 24 warnings to only 8 warnings and no errors. ✅

GermanSaracca:
Added waiting tag, might be good to wait to merge this issue until we get the e2e tests fixed.

MellyGray (approved):

PR #360, opened 2024-04-01
Feature/294 clear test datasets

### LINKED ISSUE #294: Clear test datasets before integration test

ekraffmiller:
Deletes test datasets before e2e tests, so that destroyAll() doesn't fail when trying to delete too many tests at once.

- Closes #294

Run the e2e tests, check that test datasets have been deleted after the tests run. Check that "gets the total dataset count" test in DatasetJSDataverseRepository.spec.ts runs without errors

no


MellyGray (changes requested):
We're also creating datasets in the `FileJSDataverseRepository.spec.ts` and `Collection.spec.tsx` files. We should include the `destroyAll` method there as well.

Additionally, since we're now destroying all datasets at the beginning of each test, we should consider removing all calls to the `destroyAll` method within the individual tests. Such as in `cy.wrap(DatasetHelper.destroyAll().then(() => DatasetHelper.createWithTitle(title)))`

ekraffmiller:
thanks @MellyGray, I made the changes you suggested

MellyGray (dismissed):
LGTM!

GermanSaracca:
Hi @ekraffmiller , 2 suites of tests are failing, I attach screenshots.
I removed images and volumes from docker, downloaded everything again and they are still occurring in the same way.
Screenshot 2024-04-03 at 09 22 03
Screenshot 2024-04-03 at 08 54 04
Screenshot 2024-04-03 at 09 21 01

ekraffmiller:
@GermanSaracca I tried it again locally, as well as re-running the action, and I haven't reproduced the errors. I think it must be a timing issue. One thing that may be different is the Docker settings.  A while ago, I had to increase my resource settings in Docker, in order to run other containers.  Here is what I have for Docker resources:
Screenshot 2024-04-03 at 11 57 35 AM
Can you check what you have, and maybe try running again with more memory?

GermanSaracca:
I unassign me from possible local environment problems to test this issue. Looks like a timing issue on my end.
@MellyGray @GPortas could any of you test this on your end?

GPortas (commented):
@ekraffmiller Can you please resolve the merge conflicts?

GPortas (approved):

PR #367, opened 2024-04-11
Fix/273 deaccessioned dataset unauthenticated

### LINKED ISSUE #273: FIx dataset loading logic for Deaccessioned Dataset
When a dataset is deaccessioned, it still should be visible to unauthenticated users, but only the citation and deaccession reason are viewable.
For example in JSF:
Screenshot 2024-01-04 at 1 45 21 PM

There is a bug in the SPA where  getDatasetFilesTotalDownloadSize() is called for a deaccessioned dataset and an unauthenticated user.  This API call returns an error because the user doesn't have permission to view the download sizes for a deaccessioned dataset.  Because of this error, the SPA redirects to a published version of the dataset.
Screenshot 2024-01-04 at 1 52 17 PM

We need to update the logic of the SPA so that getDatasetFilesTotalDownloadSize() and other file related API's are only called if the user has permission to view the files. For example, if the dataset is deaccessioned, but the user has edit permissions on the dataset.

ekraffmiller:
This PR fixes an issue where the Dataset page of a deaccessioned dataset is not viewable for unauthenticated users.

To recreate this error,  create dataset, publish it, then deaccession it. In JSF, when you log out, the deaccessioned dataset is still viewable to the unauthenticated user.  If you go to the SPA and try to view the dataset, you get 'Page Not Found'

- Closes #273 

Follow the steps to reproduce above.  Test that the Dataset page is still viewable when the user is logged out.

### DIFF
--- a/src/dataset/infrastructure/repositories/DatasetJSDataverseRepository.ts
+++ b/src/dataset/infrastructure/repositories/DatasetJSDataverseRepository.ts
@@ -31,6 +31,17 @@ import { DatasetDTOMapper } from '../mappers/DatasetDTOMapper'
 import { DatasetsWithCount } from '../../domain/models/DatasetsWithCount'
 
 const includeDeaccessioned = true
+type DatasetDetails = [JSDataset, string[], string, JSDatasetPermissions, JSDatasetLock[]]
+
+interface IDatasetDetails {
+  jsDataset: JSDataset
+  summaryFieldsNames: string[]
+  citation: string
+  jsDatasetPermissions: JSDatasetPermissions
+  jsDatasetLocks: JSDatasetLock[]
+  jsDatasetFilesTotalOriginalDownloadSize: number
+  jsDatasetFilesTotalArchivalDownloadSize: number
+}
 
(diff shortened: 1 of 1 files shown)


GermanSaracca (approved):
Nice fix, looks great to me. I have already tested the behavior in the JSF version and tested the behavior in this branch and everything works as expected.

The error was occurring because Promise.all failed due to the api call to downloadSizes when canEditDataset was false right?
Attached is a screenshot of how a non-authenticated user can view a deaccesioned dataset.

Screenshot 2024-04-11 at 11 15 57

ekraffmiller:
> Nice fix, looks great to me. I have already tested the behavior in the JSF version and tested the behavior in this branch and everything works as expected.
> 
> The error was occurring because Promise.all failed due to the api call to downloadSizes when canEditDataset was false right? Attached is a screenshot of how a non-authenticated user can view a deaccesioned dataset.
> 
Yes, that's right, the unauthenticated user doesn't have permission to get the downloadSizes

MellyGray (approved):
LGTM! Deaccessioned page it's working well with an unauthenticated user

PR #383, opened 2024-04-23
Feat: Add pagination total results

### LINKED ISSUE #330: Format total in table pagination results according to browser language.
The objective is to format the total number of items of the pagination result according to the preferred language in the user's browser.

total_files

For example:
- **1 to 10 of 10,000 Files** _(with 'en' locale)_
- **1 to 10 of 10.000 Files** _(with 'es' locale)_

We could use `Intl.NumberFormat` native browser api, passing undefined as language.
It has a %97.16 browser support so it is more than safe to use according to [caniuse](https://caniuse.com/?search=Intl.NumberFormat)

b-hok:
Formats the total number of items of the pagination result according to the user's preferred language. Applied to locales/en/pagination

Closes #330 

Storybook (There are additional areas that can be viewed too)
Sample URL 1: http://localhost:6006/?path=/story/sections-dataset-page-datasetfiles--no-filters

Cypress
Relating tests found in PaginationResultsInfo.spec.tsx

See Issue #330 

**Additional documentation**:

### DIFF
--- a/src/sections/shared/pagination/PaginationResultsInfo.tsx
+++ b/src/sections/shared/pagination/PaginationResultsInfo.tsx
@@ -18,8 +18,8 @@ export function PaginationResultsInfo({ paginationInfo, accumulated }: Paginatio
       accumulated === 1
         ? 'accumulated.one'
         : accumulated < paginationInfo.pageSize
-        ? 'accumulated.lessThanPageSize'
-        : 'accumulated.moreThanPageSize',
+          ? 'accumulated.lessThanPageSize'
+          : 'accumulated.moreThanPageSize',
     [accumulated, paginationInfo.pageSize]
   )
 
@@ -27,16 +27,18 @@ export function PaginationResultsInfo({ paginationInfo, accumulated }: Paginatio
     <span className={styles.results}>
       {typeof accumulated === 'number'
         ? t(defineLocale(accumulated), {
-            accumulated: accumulated,
-            count: paginationInfo.totalItems,
-            item: paginationInfo.itemName
-          })
+          accumulated: accumulated,
+          count: paginationInfo.totalItems,
+          countFormat: new Intl.NumberFormat().format(paginationInfo.totalItems),
+          item: paginationInfo.itemName
+        })
         : t('results', {
-            start: paginationInfo.pageStartItem,
-            end: paginationInfo.pageEndItem,
-            item: paginationInfo.itemName,
-            count: paginationInfo.totalItems
-          })}
+          start: paginationInfo.pageStartItem,
+          end: paginationInfo.pageEndItem,
+          item: paginationInfo.itemName,
+          count: paginationInfo.totalItems,
+          countFormat: new Intl.NumberFormat().format(paginationInfo.totalItems)
+        })}
     </span>
   )
 }
--- a/src/sections/shared/pagination/PaginationResultsInfo.tsx
+++ b/src/sections/shared/pagination/PaginationResultsInfo.tsx
@@ -1,4 +1,4 @@
-import { useCallback } from 'react'
+import { useCallback, useMemo } from 'react'
 import styles from './Pagination.module.scss'
 import { PaginationInfo } from '../../../shared/pagination/domain/models/PaginationInfo'
 import { useTranslation } from 'react-i18next'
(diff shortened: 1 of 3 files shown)


src/sections/shared/pagination/PaginationResultsInfo.tsx
```diff
@@ -18,25 +18,27 @@ export function PaginationResultsInfo({ paginationInfo, accumulated }: Paginatio
...
     <span className={styles.results}>
       {typeof accumulated === 'number'
         ? t(defineLocale(accumulated), {
-            accumulated: accumulated,
-            count: paginationInfo.totalItems,
-            item: paginationInfo.itemName
-          })
+          accumulated: accumulated,
+          count: paginationInfo.totalItems,
+          countFormat: new Intl.NumberFormat().format(paginationInfo.totalItems),
```
GermanSaracca, line 32: Perhaps instead of countFormat naming we could use formattedCount instead?

src/sections/shared/pagination/PaginationResultsInfo.tsx
```diff
@@ -18,25 +18,27 @@ export function PaginationResultsInfo({ paginationInfo, accumulated }: Paginatio
...
-            start: paginationInfo.pageStartItem,
-            end: paginationInfo.pageEndItem,
-            item: paginationInfo.itemName,
-            count: paginationInfo.totalItems
-          })}
+          start: paginationInfo.pageStartItem,
+          end: paginationInfo.pageEndItem,
+          item: paginationInfo.itemName,
+          count: paginationInfo.totalItems,
+          countFormat: new Intl.NumberFormat().format(paginationInfo.totalItems)
```
GermanSaracca, line 40: We are formatting `totalItems` twice, perhaps it would be cleaner to store this `formattedCount` in a variable, outside of the component's return function.
We could also use a `useMemo` hook to recalculate that value only if `totalItems` changes.
What do you think ?

tests/component/sections/shared/pagination/PaginationResultsInfo.spec.tsx
```diff
@@ -53,4 +53,15 @@ describe('PaginationResultsInfo', () => {
...
-})
+
+  it('shows the correct formatted results when accumulated prop passed and results are more than page size', () => {
+    cy.customMount(
+      <PaginationResultsInfo
+        paginationInfo={new PaginationInfo<DatasetPaginationInfo>(1, 100, 1500)}
+        accumulated={500}
+      />
+    )
+    cy.findByText('500 of 1,500 Items seen').should('exist')
```
GermanSaracca, line 64: Maybe we could format again the expected total, so we don't hardcode the translated total.
Something like:
```
const formattedTotalCount = new Intl.NumberFormat().format(1500)

cy.findByText(`500 of ${formattedTotalCount} Items seen`).should('exist')
```

GermanSaracca (changes requested):
Hi @b-hok, many thanks for your contribution !
It looks great overall, I just have a few comments to share with you.

src/sections/shared/pagination/PaginationResultsInfo.tsx
b-hok: This is a great idea, I agree.

tests/component/sections/shared/pagination/PaginationResultsInfo.spec.tsx
b-hok: Made a slight tweak and added two variables: `expectedFormattedTotalCount` and `totalCount` to help with the hardcoding.

tests/component/sections/shared/pagination/PaginationResultsInfo.spec.tsx
```diff
@@ -53,4 +53,17 @@ describe('PaginationResultsInfo', () => {
     )
     cy.findByText('10 of 15 Items seen').should('exist')
   })
-})
+
+  it('shows the correct formatted results when accumulated prop passed and results are more than page size', () => {
+    const totalCount = 1500
+    const expectedFormattedTotalCount = '1,500'
```
GermanSaracca, line 59: What do you think that instead of hardcoding the expected result, we use Intl here too?
So we are not hardcoding the expected to any specific language.
`const expectedFormattedTotalCount = new Intl.NumberFormat().format(totalCount)`

b-hok: The two approaches work. 

If it is hard-coded, it is easier to see what is the expected result for that specific language. For en, we know for sure if it "1,500" and we know there is a need for a comma.

If it is not hard-coded, it is slightly easier to reconfigure and run.

I made the change already and I can change it again to whichever is most preferred.

GermanSaracca (approved):
LGTM now! Approved!
We will move to QA this issue and merge it later if everything is working as expected!
Many thanks for your contribution to Dataverse Frontend! 🙌

tests/component/sections/shared/pagination/PaginationResultsInfo.spec.tsx
GermanSaracca: Both valid points, thanks for your contribution!

GPortas (approved):

PR #395, opened 2024-05-13
Multiple dynamic fields - Create Dataset Form

### LINKED ISSUE #376: Add dynamic multiple field types to Create Dataset Form
Update the Create Dataset form so it includes multiple field types, by integrating the logic in the [AuthorFormGroup](https://github.com/IQSS/dataverse-frontend/blob/develop/src/sections/create-dataset/AuthorFormGroup.tsx) component

GermanSaracca:
Adds the ability to add multiple values to a field.

Closes #376

- New **_Select Multiple_** component integrated in the form to replace the multiple checkbox.
- _**FormGroupWithMultipleFields**_  is no longer responsible for adding or removing elements, that is taken care of by the consumer logic (SPA).
- I need to remove Form Group `required` and `withinMultipleFields` props being cloned to all childrens inside because did not work correctly when we had nested elements to clone within the Forms Group (e.g. adding props to all elements unnecessarily).
- I removed `withinMultipleFields` props from form elements and with it the `<FormElementLayout/>`, so now the design system is not responsible for rendering the form group as a column or as a row.
In some cases within the Create Dataset form when working on this issue I found some complications when working with so many dynamic fields.

I think it's pretty useful to just pass a prop to render the form group one way or another, but I think the use of `react.cloneElement` was being counter-intuitive in this case.
Perhaps we could see in a separate issue a way to accomplish the same thing but without the use of `cloneElement`.

### Step 1: Run the Development Environment
1. Execute `npm i`.
3. Return with `cd ../../`.

### Step 2: Test the Create Dataset Form
- Log in, click the create dataset button and the form fields should be displayed successfully.
- Add or remove a field by clicking the buttons to the right of it.
- Complete all required fields and save the form, everything should keep working as usual.
- Try closing the accordion and clicking the Save button without filling any field, the accordion with errors should open and scroll and focus to the first field with errors.

**Does this PR introduce a user interface change? If mockups are available, please link/include them here:
Yes.
DynamicMultipleFields

**Additional documentation**:
No.


tests/component/sections/create-dataset/CreateDataset.spec.tsx
ekraffmiller, line 1: `cy.wait(250) // Maybe not needed? `
 Yes, this can be removed.  Cypress will wait a set time (I think the default is 4 seconds) for the html to update

ekraffmiller (commented):
This looks good and works very well!  
A general comment I have is, could we MetadataBlockFormFields and the components within it to a shared folder?  I think that code will end up being used in the Edit Dataset action, and also it seems like code that other UIs would want to reuse.  Also creating stories and tests for these components I think would help with re-use.  What do you think?

tests/component/sections/create-dataset/CreateDataset.spec.tsx
GermanSaracca: haha at least I left a comment with my doubt 🤣, both waits removed

GermanSaracca:
> This looks good and works very well! A general comment I have is, could we MetadataBlockFormFields and the components within it to a shared folder? I think that code will end up being used in the Edit Dataset action, and also it seems like code that other UIs would want to reuse. Also creating stories and tests for these components I think would help with re-use. What do you think?

Hi @ekraffmiller I thought the same thing at the time, but I think it is not yet ready to be in a shared component, it will have to be more generic yet, but indeed that should be the direction we have to follow, maybe when the time comes to make the edit form we could simply make the necessary changes to make it fully functional for both create and edit and then put it in shared folder, what do you think?

ekraffmiller:
> > This looks good and works very well! A general comment I have is, could we MetadataBlockFormFields and the components within it to a shared folder? I think that code will end up being used in the Edit Dataset action, and also it seems like code that other UIs would want to reuse. Also creating stories and tests for these components I think would help with re-use. What do you think?
> 
> Hi @ekraffmiller I thought the same thing at the time, but I think it is not yet ready to be in a shared component, it will have to be more generic yet, but indeed that should be the direction we have to follow, maybe when the time comes to make the edit form we could simply make the necessary changes to make it fully functional for both create and edit and then put it in shared folder, what do you think?

Ok, That makes sense!  We will have a better idea how to refactor it when we do the edit page.

ekraffmiller:
I noticed the e2e tests are failing, maybe we can merge the latest from develop when the race condition PR is merged, to make sure the tests are ok.

tests/component/sections/create-dataset/CreateDataset.spec.tsx
ekraffmiller: lol thanks for the comment, it caught my attention :)

GermanSaracca:
> I noticed the e2e tests are failing, maybe we can merge the latest from develop when the race condition PR is merged, to make sure the tests are ok.

Yes, I think that will be the safest thing to do.

GermanSaracca:
@ekraffmiller all e2e tests are passing, I only update the test for the create dataset form to select an option from the select multiple field.

ekraffmiller (approved):
Looks good. approved!

PR #397, opened 2024-05-16
File Uploader

### LINKED ISSUE #366: Add File Upload Page UI components
Add File Upload Page UI components

Depends on https://github.com/IQSS/dataverse-frontend/issues/365

Features:

1. File upload box (without dropbox option)

## Mock

Component 138 (1)

All users

## What inspired the request?
Frontend meeting discussion about the next features planned for Q2

None

New File Upload Page in the SPA beta

- https://github.com/IQSS/dataverse-frontend/issues/365

ErykKul:
It adds a new file upload component to frontend.

Closes #366

### DIFF
--- a/src/sections/upload-dataset-files/FileUploader.tsx
+++ b/src/sections/upload-dataset-files/FileUploader.tsx
@@ -130,7 +130,8 @@ export function FileUploader({
           onDrop={handleDrop}
           onDragOver={handleDragOver}
           onDragEnter={handleDragEnter}
-          onDragLeave={handleDragLeave}>
+          onDragLeave={handleDragLeave}
+          data-testid="drag-and-drop">
           <div>
             <input
               ref={inputRef}
--- a/tests/component/sections/upload-dataset-files/UploadDatasetFiles.spec.tsx
+++ b/tests/component/sections/upload-dataset-files/UploadDatasetFiles.spec.tsx
@@ -71,13 +71,13 @@ describe('UploadDatasetFiles', () => {
 
     mountWithDataset(<UploadDatasetFiles fileRepository={new FileMockRepository()} />, testDataset)
 
-    const dnd = cy.findByText('Drag and drop files here.')
-    dnd.should('exist')
-    const uploadFile = (fileName: string, contents: unknown) => {
-      dnd.selectFile({ fileName: fileName, contents: contents }, { action: 'drag-drop' })
-    }
+    cy.findByTestId('drag-and-drop').as('dnd')
+    cy.get('@dnd').should('exist')
 
-    uploadFile('users1.json', [{ name: 'John Doe the 1st' }])
+    cy.get('@dnd').selectFile(
+      { fileName: 'users1.json', contents: [{ name: 'John Doe the 1st' }] },
+      { action: 'drag-drop' }
+    )
     cy.findByText('users1.json').should('exist')
     cy.findByTitle('Cancel upload').should('exist')
     cy.findByRole('progressbar').should('exist')
@@ -89,15 +89,21 @@ describe('UploadDatasetFiles', () => {
 
     mountWithDataset(<UploadDatasetFiles fileRepository={new FileMockRepository()} />, testDataset)
 
-    const dnd = cy.findByText('Drag and drop files here.')
-    dnd.should('exist')
-    const uploadFile = (fileName: string, contents: unknown) => {
-      dnd.selectFile({ fileName: fileName, contents: contents }, { action: 'drag-drop' })
-    }
+    cy.findByTestId('drag-and-drop').as('dnd')
+    cy.get('@dnd').should('exist')
 
-    uploadFile('users1.json', [{ name: 'John Doe the 1st' }])
-    uploadFile('users2.json', [{ name: 'John Doe the 2nd' }])
-    uploadFile('users3.json', [{ name: 'John Doe the 3rd' }])
+    cy.get('@dnd').selectFile(
+      { fileName: 'users1.json', contents: [{ name: 'John Doe the 1st' }] },
+      { action: 'drag-drop' }
+    )
+    cy.get('@dnd').selectFile(
+      { fileName: 'users2.json', contents: [{ name: 'John Doe the 2nd' }] },
+      { action: 'drag-drop' }
+    )
+    cy.get('@dnd').selectFile(
+      { fileName: 'users3.json', contents: [{ name: 'John Doe the 3rd' }] },
+      { action: 'drag-drop' }
+    )
     cy.findAllByTitle('Cancel upload').first().parent().click()
     cy.findByText('users1.json').should('not.exist')
     cy.findByText('users2.json').should('exist')
@@ -138,13 +144,13 @@ describe('UploadDatasetFiles', () => {
       testDataset
     )
 
-    const dnd = cy.findByText('Drag and drop files here.')
-    dnd.should('exist')
-    const uploadFile = (fileName: string, contents: unknown) => {
-      dnd.selectFile({ fileName: fileName, contents: contents }, { action: 'drag-drop' })
-    }
+    cy.findByTestId('drag-and-drop').as('dnd')
+    cy.get('@dnd').should('exist')
 
-    uploadFile('users1.json', [{ name: 'John Doe the 1st' }])
+    cy.get('@dnd').selectFile(
+      { fileName: 'users1.json', contents: [{ name: 'John Doe the 1st' }] },
+      { action: 'drag-drop' }
+    )
     cy.findByText('users1.json').should('exist') //.should('have.class', 'failed')
   })
 
@@ -153,15 +159,21 @@ describe('UploadDatasetFiles', () => {
 
     mountWithDataset(<UploadDatasetFiles fileRepository={new FileMockRepository()} />, testDataset)
 
-    const dnd = cy.findByText('Drag and drop files here.')
-    dnd.should('exist')
-    const uploadFile = (fileName: string, contents: unknown) => {
-      dnd.selectFile({ fileName: fileName, contents: contents }, { action: 'drag-drop' })
-    }
+    cy.findByTestId('drag-and-drop').as('dnd')
+    cy.get('@dnd').should('exist')
 
-    uploadFile('users1.json', [{ name: 'John Doe the 1st' }])
-    uploadFile('users1.json', [{ name: 'John Doe the 1st' }])
-    uploadFile('users3.json', [{ name: 'John Doe the 3rd' }])
+    cy.get('@dnd').selectFile(
+      { fileName: 'users1.json', contents: [{ name: 'John Doe the 1st' }] },
+      { action: 'drag-drop' }
+    )
+    cy.get('@dnd').selectFile(
+      { fileName: 'users1.json', contents: [{ name: 'John Doe the 1st' }] },
+      { action: 'drag-drop' }
+    )
+    cy.get('@dnd').selectFile(
+      { fileName: 'users3.json', contents: [{ name: 'John Doe the 3rd' }] },
+      { action: 'drag-drop' }
+    )
     cy.findByText('users3.json').should('exist')
     cy.findByText('users1.json').should('exist')
     cy.findAllByTitle('Cancel upload').should('have.length', 2)
@@ -170,8 +182,14 @@ describe('UploadDatasetFiles', () => {
     // wait for upload to finish
     cy.findByText('users3.json').should('not.exist')
     cy.findByText('users1.json').should('not.exist')
-    uploadFile('users1.json', [{ name: 'John Doe the 1st' }])
-    uploadFile('users3.json', [{ name: 'John Doe the 3rd' }])
+    cy.get('@dnd').selectFile(
+      { fileName: 'users1.json', contents: [{ name: 'John Doe the 1st' }] },
+      { action: 'drag-drop' }
+    )
+    cy.get('@dnd').selectFile(
+      { fileName: 'users3.json', contents: [{ name: 'John Doe the 3rd' }] },
+      { action: 'drag-drop' }
+    )
     cy.findByText('users3.json').should('not.exist')
     cy.findByText('users1.json').should('not.exist')
   })
@@ -181,15 +199,21 @@ describe('UploadDatasetFiles', () => {
 
     mountWithDataset(<UploadDatasetFiles fileRepository={new FileMockRepository()} />, testDataset)
 
-    const dnd = cy.findByText('Drag and drop files here.')
-    dnd.should('exist')
-    const uploadFile = (fileName: string, contents: unknown) => {
-      dnd.selectFile({ fileName: fileName, contents: contents }, { action: 'drag-drop' })
-    }
+    cy.findByTestId('drag-and-drop').as('dnd')
+    cy.get('@dnd').should('exist')
 
-    uploadFile('users1.json', [{ name: 'John Doe the 1st' }])
-    uploadFile('users1.json', [{ name: 'John Doe the 1st' }])
-    uploadFile('users3.json', [{ name: 'John Doe the 3rd' }])
+    cy.get('@dnd').selectFile(
+      { fileName: 'users1.json', contents: [{ name: 'John Doe the 1st' }] },
+      { action: 'drag-drop' }
+    )
+    cy.get('@dnd').selectFile(
+      { fileName: 'users1.json', contents: [{ name: 'John Doe the 1st' }] },
+      { action: 'drag-drop' }
+    )
+    cy.get('@dnd').selectFile(
+      { fileName: 'users3.json', contents: [{ name: 'John Doe the 3rd' }] },
+      { action: 'drag-drop' }
+    )
     cy.findByText('users3.json').should('exist')
     cy.findByText('users1.json').should('exist')
     cy.findAllByTitle('Cancel upload').should('have.length', 2)
@@ -202,11 +226,8 @@ describe('UploadDatasetFiles', () => {
 
     mountWithDataset(<UploadDatasetFiles fileRepository={new FileMockRepository()} />, testDataset)
 
-    const dnd = cy.findByText('Drag and drop files here.')
-    dnd.should('exist')
-    const uploadFile = (fileName: string, contents: unknown) => {
-      dnd.selectFile({ fileName: fileName, contents: contents }, { action: 'drag-drop' })
-    }
+    cy.findByTestId('drag-and-drop').as('dnd')
+    cy.get('@dnd').should('exist')
 
     const filenames: string[] = [
       'users1.json',
@@ -221,7 +242,10 @@ describe('UploadDatasetFiles', () => {
       'users10.json'
     ]
     filenames.forEach((element) => {
-      uploadFile(element, [{ name: 'John Doe' }])
+      cy.get('@dnd').selectFile(
+        { fileName: element, contents: [{ name: 'John Doe' }] },
+        { action: 'drag-drop' }
+      )
     })
     cy.findAllByTitle('Cancel upload').should('have.length', 10)
     cy.findAllByRole('progressbar').should('have.length', 6)
@@ -239,7 +263,10 @@ describe('UploadDatasetFiles', () => {
       'users20.json'
     ]
     filenames2.forEach((element) => {
-      uploadFile(element, [{ name: 'John Doe' }])
+      cy.get('@dnd').selectFile(
+        { fileName: element, contents: [{ name: 'John Doe' }] },
+        { action: 'drag-drop' }
+      )
     })
     cy.findByText('users20.json').should('exist')
     cy.findAllByRole('progressbar').should('have.length', 6)


ErykKul:
Work in progress. This PR is already early, before the work is done, for getting early feedback, since it is my first pull request on the new SPA.

packages/design-system/src/lib/components/file-uploader/FileUploader.tsx
```diff
@@ -0,0 +1,178 @@
...
+                              : styles.cell
+                          }>
+                          {file.webkitRelativePath}
+                          {file.name}
+                        </div>
+                        <div className={styles.cell}>
+                          {progress.get(file.webkitRelativePath + file.name)?.fileSizeString}
+                        </div>
+                        <div className={styles.cell}>
+                          <ProgressBar
```
GermanSaracca, line 152: The bar progress is going back and forward, I am testing it from a Mac on Chrome Version 125.0.6422.60 (Official Build) (arm64) if it helps.

GermanSaracca (commented):
Hi erik! thanks for showing me your progress on this important feature for the new SPA!
I left you some comments with things that I found that could be improved.

Something very relevant that we have discussed with the frontend team is that this component should not live in the design system but directly in the SPA.
The design system is an external package, a library of components reusable and reused throughout the SPA itself.
It would be best to create this component initially within the SPA, specifically in the `src/sections/upload-dataset-files` folder, once we have it fully functional and stable, and if we see that it is reused in several places in the SPA, then we could analyse the option of adding it to the design system.

Please feel free to let us know if you have any doubts or questions through the slack channel or wherever you feel necessary, thank you very much for contributing to the new SPA!

packages/design-system/src/lib/components/file-uploader/FileUploader.tsx
ErykKul: I made it move at random, just for testing, by assigning a progress value at random between 0 and 100. In real life, the progress should be a percentage of the bytes uploaded, making the bar only move forward.

ErykKul:
@GermanSaracca 
Thanks for the feedback! I will implement all your suggestions, these are great! I do not have any preference to keep the component in the storybook, I just did not know where it would be at home. Furthermore, I will move it to the SPA source, and it can stay there, unless someone requests it to be moved to storybook.

packages/design-system/src/lib/components/file-uploader/FileUploader.tsx
```diff
@@ -0,0 +1,178 @@
...
+      <Card.Header>
+        <Button variant="secondary" onClick={() => (inputRef.current! as HTMLInputElement).click()}>
+          <Plus></Plus> {selectText}
+        </Button>
+      </Card.Header>
+      <Card.Body>
+        <div
+          className={styles.fileuploader}
+          onDrop={handleDrag}
+          onDragOver={(event) => event.preventDefault()}>
```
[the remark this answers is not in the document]
ErykKul: I changed the background color, we can decide on a specific color, etc., later. Now I used the info box color on drag enter, drag leave and drag over.

GermanSaracca:
@ErykKul  There are two storybooks, one for the Design System components only and one for the SPA.
You should create the story about the FileUploader in `src/stories/upload-dataset-files/UploadFiles.stories.tsx`
Many thanks!

ErykKul:
There is already one that looks fine:
image

Do you mean an extra one, for the file upload element only?

GermanSaracca:
> Do you mean an extra one, for the file upload element only?

A story for the UploadDatasetFile page that includes all the elements inside it is fine, as the page gets bigger it may be useful to also create stories for some components separately, but it is fine with one story for now. 👍🏼

packages/design-system/src/lib/stories/file-uploader/FileUploader.stories.tsx
```diff
@@ -0,0 +1,39 @@
...
+  tags: ['autodocs']
+}
+
+export default meta
+type Story = StoryObj<typeof FileUploader>
+
+export const Default: Story = {
+  render: () => (
+    <FileUploader
+      upload={function (files: File[]): Map<string, FileUploadState> {
```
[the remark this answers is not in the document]
ErykKul: It went in a slightly different direction, I moved the state one component up. It will be a shared state between two components; there is also an uploaded files list at the bottom with a save button (not implemented yet). Anyway, it is more finished now and should be clearer where I was going with this. I tested it with several thousands of files at once, it handles it well. Should be more than enough for the intended use case. We might want to add something to limit how many files are uploaded at once, if not implemented yet in the jsdv library.

ErykKul:
@GermanSaracca 
I cleaned up the code and implemented the use case and the mock of the upload file service. I still need to add the uploaded files list at the bottom, and implement translations for the texts, etc., but it should be more clear now where it is going. If you find a moment, can you take a look at the code as it is now? Thanks!

GermanSaracca:
Hi @ErykKul , the scope of the issue is just to create the FileUploader component (dropzone) and a callback to get the files processed from it, not the list of files for now ( that will be added in a near future)
We can keep this simple for now and in another iteration actually connect to the use case of the client-javascript repo to actually upload the files.
@ekraffmiller @GPortas what do you think?

GermanSaracca:
> If you find a moment, can you take a look at the code as it is now? Thanks!

Sure, I will try to take a look at it tomorrow. Many thanks!

ErykKul:
I added the semaphore (with limit 6, I think that it depends on browser and number of connections allowed, we can choose a different value at some point) and the translations. I think it is ready for review then (assuming we are keeping the scope small for this PR). If I remember correctly, we will still need one more callback for the storage identifier, this one can be easily added in the future using the `FileUploadTools` in [FileUploadState.ts](https://github.com/IQSS/dataverse-frontend/pull/397/files#diff-1cb1f0fbb692bfb3e404dd9d0c3eea6e073ecffeeff3302a9571a9d07ffee457).

src/sections/upload-dataset-files/FileUploader.tsx
```diff
@@ -0,0 +1,187 @@
+import { ChangeEventHandler, DragEventHandler, useEffect, useState } from 'react'
+import { ProgressBar, Card } from 'react-bootstrap'
```
GermanSaracca, line 2: We try not to use react-bootstrap components directly from SPA, we always try to import them from our design system.
So if you need a react-bootstrap component that is not yet in the design system, you should re-export it from the design system, apply the props interface you see necessary so we don't depend on the library interface itself, add a story to storybook, a basic unit test and document the new addition in the CHANGELOG.md under # Non Published Changes.
This way if in the future we change the component library used in the design system we are not so coupled to it from the SPA.
If you need further explanation on this, please let me know or we can have a quick pair progamming.

src/sections/upload-dataset-files/UploadDatasetFiles.tsx
```diff
@@ -1,19 +1,104 @@
...
+  const [semaphore, setSemaphore] = useState(new Set<string>())
+
+  const timeout = (delay: number) => new Promise((res) => setTimeout(res, delay))
+  const limit = 6
+
+  const acquireSemaphore = async (file: File) => {
+    const key = FileUploadTools.key(file)
+    setSemaphore((x) => (x.size >= limit ? x : x.add(key)))
+    while (!semaphore.has(key)) {
+      await timeout(500)
```
GermanSaracca, line 32: Why do you need to manually delay this process?

src/sections/upload-dataset-files/UploadDatasetFiles.tsx
```diff
@@ -1,19 +1,104 @@
...
   const { dataset, isLoading } = useDataset()
+  const [fileUploaderState, setState] = useState(FileUploadTools.createNewState([]))
+  const [uploadingToCancelMap, setUploadingToCancelMap] = useState(new Map<string, () => void>())
+  const [uploadFinished, setUploadFinished] = useState(new Set<string>())
+  const [semaphore, setSemaphore] = useState(new Set<string>())
+
+  const timeout = (delay: number) => new Promise((res) => setTimeout(res, delay))
+  const limit = 6
+
+  const acquireSemaphore = async (file: File) => {
```
GermanSaracca, line 28: Could you explain me how this semaphore technique works and why we need it? Looks really nice!

GermanSaracca (changes requested):
Hi @ErykKul , I think it looks good, but also that there are a lot of things going on that I'm not sure if we need them this point, maybe you have a better idea of how this works in the JSF version and that's why there are so many things involved in this PR like, semaphores, progress tracking, abort controllers( @GPortas are we planning to add the ability to abort the upload right now and also keep track of the progress of each file from the use case? )

One thing I have noticed is that when a file is cancelled, all files are, but that part is more for the QA person, I will stick to code review.

Could you add some unit/integration tests ? that way I will understand more what each part does and what is expected, we try to keep each component tested and we have a 95% test coverage threshold.

src/sections/upload-dataset-files/UploadDatasetFiles.tsx
```diff
@@ -1,19 +1,104 @@
...
 }
 
-export const UploadDatasetFiles = ({
-  fileRepository: _fileRepository
-}: UploadDatasetFilesProps) => {
+export const UploadDatasetFiles = ({ fileRepository: fileRepository }: UploadDatasetFilesProps) => {
+  const { t } = useTranslation('files')
   const { setIsLoading } = useLoading()
   const { dataset, isLoading } = useDataset()
+  const [fileUploaderState, setState] = useState(FileUploadTools.createNewState([]))
```
GermanSaracca, line 20: Notice that you can use react useReducer when you have lots of states and logic going on.
That way you can keep the state synced in one place and read it from one place also.
Maybe you don't need it, just a comment :).

https://react.dev/reference/react/useReducer

Here is an example in the [selectMultipleReducer](https://github.com/IQSS/dataverse-frontend/blob/develop/packages/design-system/src/lib/components/select-multiple/selectMultipleReducer.ts)

src/sections/upload-dataset-files/UploadDatasetFiles.tsx
ErykKul: I could have used here events and listeners, or some other publish-subscribe thing. But it would make the code more complex. I looked for some examples I could find online, but did not like them. Keeping the while loop without delay makes one CPU core go 100%, the delay solves it and the upload feels responsive enough. There might be a better solution for this, I did not find one yet...

src/sections/upload-dataset-files/UploadDatasetFiles.tsx
ErykKul: I hope it clarifies it somewhat, if not we could discuss it in a meeting. I also do not mind deleting it, or removing some other parts of the code, if it is too much, etc.

src/sections/upload-dataset-files/UploadDatasetFiles.tsx
ErykKul: I did not know this. It looks very similar to what I did. It is my first time I did write some code with React, so it is hard to tell if the reducer would be better here. The function calls look more natural to me in the callbacks as I did in my code, and the reducer looks more natural for buttons, etc., but it might be just me, because I do not know React very well... I will leave it like it is for now, we might want to delete some of my code, or change how the state works, what is in it, etc.

src/sections/upload-dataset-files/UploadDatasetFiles.tsx
GermanSaracca: Thank you very much for the explanation. Taking care of CPU performance and controlling processes in such a granular way is the way to go! 
I think it's great, you're looking at the big picture and we're going to need all those features!
👏🏼 👏🏼

ErykKul:
I am glad you noticed the cancel problem, it is better to know it early, and not wait for the QA. I think I might have broken something at some point, I am sure it worked. Maybe there is a problem with my mock implementation, and it is doing too much at once (it listens to cancels, updates the progress and notifies when upload finished).

As for the features, if we plan to add them at some point, it is easier to foresee them in the design now, than try to add it later. We can just disable them for now and use only the callbacks in the use case (failed and done). Not having the cancel working only means that the file is uploaded to the storage, but will not be added with the addFiles API call. And the progress is just nice-to-have, it can be hidden.

I did not inspire this code on JSF implementation (only the look is copied, the implementation is new), I just implemented the features I thought were necessary for a good user experience.

GermanSaracca:
Glad to help!  😄 
I think that all the features like abort controller, progress etc are going to be necessary, I think it's good if you keep them.
What I really would try now is that all the parts and flows have their respective tests, I don't mean e2e test because those will be done at the moment of integrating with the use case, but yes the unit/integration tests with mocked data.

GermanSaracca:
Hi @ErykKul , try passing an aria-label attribute as a prop, should fix it, even better you could detect if not aria-label prop has been passed you could set one as default.

```
export function ProgressBar({
  now,
  'aria-label': ariaLabel = 'progress bar',
  ...rest
}: ProgressBarProps) {
  return <CProgressBarBS now={now} aria-label={ariaLabel} {...rest} />
}
```
This way, if the consumer of this component forgot to add an aria-label attribute, 'progress bar' will be the default, otherwise the one received as prop will be used.

ErykKul:
@GermanSaracca 
The fix for the accessibly problem did not work. Also, the test `'renders skeleton while loading'` is failing for the upload file section, I could not figure out why. I do not think I can fix those two issues by myself...
I did add the file upload test, that one does work. I have also addressed the other requests for changes. I will move this PR to ready for review.

ErykKul:
I merged the develop branch, but it broke other tests:
```  Running:  sections/upload-dataset-files/UploadDatasetFiles.spec.tsx                     (11 of 98)


  UploadDatasetFiles
    1) renders the breadcrumbs
    ✓ renders skeleton while loading (13ms)
    ✓ renders page not found when dataset is null (30ms)
    2) renders the breadcrumbs
    ✓ renders the file uploader (63ms)


  3 passing (9s)
  2 failing

  1) UploadDatasetFiles
       renders the breadcrumbs:
     AssertionError: Timed out retrying after 4000ms: Cannot read properties of undefined (reading 'DEBUG_PRINT_LIMIT')
      at Context.<anonymous> (/home/eryk/projects/dataverse-frontend/tests/component/sections/upload-dataset-files/UploadDatasetFiles.spec.tsx:35:34)
      at runnable.fn (http://localhost:5173/__cypress/runner/cypress_runner.js:160859:19)
      at callFn (http://localhost:5173/__cypress/runner/cypress_runner.js:114392:21)
      at <unknown> (http://localhost:5173/__cypress/runner/cypress_runner.js:167847:30)
      at PassThroughHandlerContext.finallyHandler (http://localhost:5173/__cypress/runner/cypress_runner.js:15298:23)
      at PassThroughHandlerContext.tryCatcher (http://localhost:5173/__cypress/runner/cypress_runner.js:18744:23)
      at Promise._settlePromiseFromHandler (http://localhost:5173/__cypress/runner/cypress_runner.js:16679:31)

  2) UploadDatasetFiles
       renders the breadcrumbs:
     AssertionError: Timed out retrying after 4000ms: Cannot read properties of undefined (reading 'DEBUG_PRINT_LIMIT')
      at Context.<anonymous> (/home/eryk/projects/dataverse-frontend/tests/component/sections/upload-dataset-files/UploadDatasetFiles.spec.tsx:60:51)
      at runnable.fn (http://localhost:5173/__cypress/runner/cypress_runner.js:160859:19)
      at callFn (http://localhost:5173/__cypress/runner/cypress_runner.js:114392:21)
      at <unknown> (http://localhost:5173/__cypress/runner/cypress_runner.js:167847:30)
      at PassThroughHandlerContext.finallyHandler (http://localhost:5173/__cypress/runner/cypress_runner.js:15298:23)
      at PassThroughHandlerContext.tryCatcher (http://localhost:5173/__cypress/runner/cypress_runner.js:18744:23)
      at Promise._settlePromiseFromHandler (http://localhost:5173/__cypress/runner/cypress_runner.js:16679:31)




  (Results)

  ┌────────────────────────────────────────────────────────────────────────────────────────────────┐
  │ Tests:        5                                                                                │
  │ Passing:      3                                                                                │
  │ Failing:      2                                                                                │
  │ Pending:      0                                                                                │
  │ Skipped:      0                                                                                │
  │ Screenshots:  2                                                                                │
  │ Video:        false                                                                            │
  │ Duration:     8 seconds                                                                        │
  │ Spec Ran:     sections/upload-dataset-files/UploadDatasetFiles.spec.tsx                        │
  └────────────────────────────────────────────────────────────────────────────────────────────────┘```

**Comment by ErykKul** [comment]
It messed up the merge. I done some fixing, and I am rerunning the tests. I think it should be fine now, except for the accessibility thing.

**Comment by ErykKul** [comment]
One test is still failing, it does not look like it is related to this PR:
```
  ├────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ ✖  sections/file/file-metadata/FileMet      00:05       29       28        1        -        - │
  │    adata.spec.tsx                                                                              │
  ├────────────────────────────────────────────────────────────────────────────────────────────────┤
```

**Comment by ErykKul** [comment]
I got it working again. I will leave it for now. The two remaining problems are the failing file metadata test (that might not be related to this PR, not sure) and the accessibility test for progress bar.

**Review by GermanSaracca — CHANGES_REQUESTED** [review]
Hi @ErykKul , looks good! Just a few observations and things that could be improved

File: src/sections/upload-dataset-files/FileUploader.module.scss [hunk 8]
```diff
@@ -0,0 +1,51 @@
+@import "node_modules/@iqss/dataverse-design-system/src/lib/assets/styles/design-tokens/colors.module";
+
+.file_uploader {
+    min-height: 5em;
+}
+
+.files {
```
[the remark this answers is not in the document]
ErykKul: I went for grid layout. Looks the same but is not a table :-)

ErykKul:
@GermanSaracca,

I implemented most of your changes, they were easy to fix as your comments were very detailed. Thanks! However, one thing remained unresolved: not using constants or vars in cypress tests. I looked into it, and it does not look like something very easy to do. I am not sure how important is this, maybe it is fine like this, and we can move this PR to QA?

Kind regards,
Eryk

ErykKul:
I have extracted the uploadFile function in the UploadDatasetFiles test. If there is a better way of simulating the drag-and-drop, that does not rely on the `dnd` constant, it can be swapped later.

public/locales/en/uploadDatasetFiles.json
```diff
@@ -1,3 +1,6 @@
 {
-  "breadcrumbActionItem": "Upload Files"
+  "breadcrumbActionItem": "Upload Files",
+  "cancel": "Cancel upload",
+  "info": "Drag and drop files here.",
+  "select": "Select Files to Add"
```
pdurbin, line 5: Should these all be "files" (lowercase) or "Files" (Title Case)?

Right now it's a mix.

pdurbin (commented):
Just a comment on "files" vs. "Files".

public/locales/en/uploadDatasetFiles.json
ErykKul: I replaced the Title Case with lower case (except the first letter of the first word). I think it is better now.

ErykKul:
After changing the case from Title Case to Regular case, some tests broke. I wonder if there is a way in the tests to use translations? I did fix the test, but it is awkward that I needed to do that. I can see how fragile these tests are.

ErykKul:
Thanks! I will adapt the tests as you suggested and move it to ready for QA.

GermanSaracca (approved):

GPortas (commented):
@ErykKul The component looks and works good. I only encountered a couple of issues in Storybook. The first is that sometimes the progress bar reverts to a previous state. Is this related to how the mocked progress is calculated for Storybook, or is it a problem with the component? Additionally, I can only test uploading a file once; it would be nice to be able to do it multiple times. I understand this is because the component's state isn't reset after the first upload? Here is a recording of these issues:

https://github.com/IQSS/dataverse-frontend/assets/17010447/3a60c407-351d-49fe-84a8-7e61eb53b9d3

FYI - As the next steps after this PR, we will need to integrate these changes with the new js-dataverse use cases that I have implemented. I have created an issue for this: https://github.com/IQSS/dataverse-frontend/issues/417

ErykKul:
@GPortas 
This component is only the first part of upload page. The uploaded files suppose to get listed at the bottom of the page with a save button:

image

Also, this feature is not yet implemented:

image

For now, the duplicate files are ignored. This is what you experienced. You can keep dropping the files, they get added to the state, that is not shown yet as a file list at the bottom. Clinking on the save button should reset the state (and finalize the upload by doing the [API call to add the uploaded to s3 files to the dataset](https://guides.dataverse.org/en/latest/developers/s3-direct-upload-api.html#adding-the-uploaded-file-to-the-dataset))

If that would help, I could already add the save button that resets the state (in mock repo, the real one would also finalize the upload).

We discussed shortly with @GermanSaracca what would be the scope of this PR, we decided to do only the upload component for now, and implement the file list, save button, and maybe add more features later. Nevertheless, I can add some features now, if needed. Also, it would be nice to work together on merging the use-cases.

GPortas (approved):
@ErykKul No problem. I believe these points will be better resolved when integrating the js-dataverse use cases. In fact, there is a use case directly related to https://guides.dataverse.org/en/latest/developers/s3-direct-upload-api.html#adding-the-uploaded-file-to-the-dataset, so we will need to add the button during the integration phase.

So I am going to merge now and unlock https://github.com/IQSS/dataverse-frontend/issues/417

Thanks!

PR #402, opened 2024-05-20
Fix/393 Match api email validation

### LINKED ISSUE #393: Create Dataset: email validation does not match validation in the Dataverse API
Go to Create Dataset form, in Point of Contact email, enter an email address that ends with a non-tld string, for example, 'ellen@test.e' and all the other required fields

- When does this issue occur?
When submitting Create Dataset form
Create Dataset
- What happens?
It shows an Alert message about a validation error, but it doesn't say what the error is, and the email field isn't highlighted
Screenshot 2024-05-01 at 3 17 17 PM

In the Network tab, the create fails with 403 and this error message:
```
{
    "status": "ERROR",
    "message": "Validation Failed: Point of Contact E-mail ellen@test.e  is not a valid email address. (Invalid value:edu.harvard.iq.dataverse.DatasetFieldValueValue[ id=null ]).java.util.stream.ReferencePipeline$3@626c7975"
}
```

GermanSaracca:
Fixes a mismatch between the email validation of the SPA Dataset creation form and the email validation of the dataverse api.
Also improves the UX to display the alert error banner only when creations fail, with messages specific to both the client-javascript api wrapper and the dataverse api itself (from the latter extracting only the field-related part of the error and not the part of the error related to Java invalid value exceptions).

Closes #393 

Enter an invalid email (test@test.e) and check that the error message is displayed below the field.
It is a bit difficult to test the error messages in the banners as the validations are being fulfilled from the form before it is submitted.
One option could be to test locally, modifying the code so that no fields are required within the form and submitting the form, in `useDefineRules.tsx` set the required property to `false`.

The same error banner is maintained, only now when the user submits the form, if there is an error and this banner appears, it is automatically focused with an specific message about the error.

**_Error thrown from the dataverse api_** 👇
Screenshot 2024-05-20 at 11 01 07

**_Error thrown from the client-javascript api wrapper_** 👇
Screenshot 2024-05-20 at 11 03 05

**_How the error banner is focused_** 👇
_In this example the client-side validations were removed just to show the focus case._
[alert-error-behaviour.webm](https://github.com/IQSS/dataverse-frontend/assets/64560524/3f33ef1a-a518-4be7-904a-2c352a79db59)

No.

**Additional documentation**:
No.

### DIFF
--- a/src/metadata-block-info/domain/models/fieldValidations.ts
+++ b/src/metadata-block-info/domain/models/fieldValidations.ts
@@ -2,7 +2,7 @@ import { DateFormats } from './MetadataBlockInfo'
 
 export function isValidEmail(email: string): boolean {
   const EMAIL_REGEX =
-    /[a-z0-9!#$%&'*+/=?^_`{|}~-]+(?:\.[a-z0-9!#$%&'*+/=?^_`{|}~-]+)*@(?:[a-z0-9](?:[a-z0-9-]*[a-z0-9])?\.)+[a-z0-9](?:[a-z0-9-]*[a-z0-9])?/
+    /[a-z0-9!#$%&'*+/=?^_`{|}~-]+(?:\.[a-z0-9!#$%&'*+/=?^_`{|}~-]+)*@(?:[a-z0-9](?:[a-z0-9-]*[a-z0-9])?\.)+[a-z0-9](?:[a-z0-9-]*[a-z0-9])/
   return EMAIL_REGEX.test(email)
 }
 
@@ -75,4 +75,31 @@ export function isValidDateFormat(dateString: string, acceptedFormat?: DateForma
       yearFormatRegex.test(dateString)
     )
   }
+}
+
+/**
+ * Extracts the invalid value from the error message, removing 'Validation Failed:' and everything after '(Invalid value:'
+ * @param errorMessage
+ * @returns the invalid value or null if it can't be extracted
+ * @example
+ * getValidationFailedFieldError("Validation Failed: Point of Contact E-mail test@test.c  is not a valid email address. (Invalid value:edu.harvard.iq.dataverse.DatasetFieldValueValue[ id=null ]).java.util.stream.ReferencePipeline$3@561b5200")
+ * // returns "Point of Contact E-mail test@test.c  is not a valid email address."
+ */
+
+export function getValidationFailedFieldError(errorMessage: string): string | null {
+  const validationFailedKeyword = 'Validation Failed:'
+  const invalidValueKeyword = '(Invalid value:'
+
+  const validationFailedKeywordIndex = errorMessage.indexOf(validationFailedKeyword)
+  const invalidValueKeywordIndex = errorMessage.indexOf(invalidValueKeyword)
+
+  if (validationFailedKeywordIndex !== -1 && invalidValueKeywordIndex !== -1) {
+    const start = validationFailedKeywordIndex + validationFailedKeyword.length
+    const end = invalidValueKeywordIndex
+    const extractedValue = errorMessage.slice(start, end).trim()
+
+    return extractedValue
+  }
+
+  return null
 }
--- a/src/sections/create-dataset/CreateDataset.tsx
+++ b/src/sections/create-dataset/CreateDataset.tsx
@@ -41,7 +41,7 @@ export function CreateDataset({
     if (!isLoadingMetadataBlocksConfiguration) {
       setIsLoading(false)
     }
-  }, [isLoading, isLoadingMetadataBlocksConfiguration])
+  }, [isLoading, isLoadingMetadataBlocksConfiguration, setIsLoading])
 
   return (
     <>
--- a/src/sections/create-dataset/DatasetForm.module.scss
+++ b/src/sections/create-dataset/DatasetForm.module.scss
@@ -1,0 +1,3 @@
+.form-container {
+  scroll-margin-top: 62px;
+}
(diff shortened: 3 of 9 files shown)


ekraffmiller:
@GermanSaracca There is still a difference between the form validation and the API validation - from my testing, I think the API validation is requiring that the email address be from a list of top level domains, for example: https://data.iana.org/TLD/tlds-alpha-by-domain.txt.  

  When I try an address that is not in the TLD list, I get an error:
Screenshot 2024-05-23 at 9 23 46 PM
The API uses an apache commons validator: https://github.com/IQSS/dataverse/blob/develop/src/main/java/edu/harvard/iq/dataverse/validation/EMailValidator.java#L13
I think it is using this underlying domain validator:
https://commons.apache.org/proper/commons-validator/apidocs/org/apache/commons/validator/routines/DomainValidator.html

I did some searches and couldn't find a javascript library that does something similar, so I'm not sure the best way to handle this, what do you think?

GermanSaracca:
> @GermanSaracca There is still a difference between the form validation and the API validation - from my testing, I think the API validation is requiring that the email address be from a list of top level domains, for example: https://data.iana.org/TLD/tlds-alpha-by-domain.txt.
> 
> When I try an address that is not in the TLD list, I get an error: Screenshot 2024-05-23 at 9 23 46 PM The API uses an apache commons validator: https://github.com/IQSS/dataverse/blob/develop/src/main/java/edu/harvard/iq/dataverse/validation/EMailValidator.java#L13 I think it is using this underlying domain validator: https://commons.apache.org/proper/commons-validator/apidocs/org/apache/commons/validator/routines/DomainValidator.html
> 
> I did some searches and couldn't find a javascript library that does something similar, so I'm not sure the best way to handle this, what do you think?

Hi @ekraffmiller , your findings are totally correct, there is no Javascript library that does the same validations as the Apache library ( at least not a maintained one ).
One option, would be to hardcode all possible TLDS in the SPA  but I really don like it because they are always changing or new ones are added.
I think this is just a client side validation layer and after receiving the error we are showing in the alert so I think it is enough for the user to know what is happening. What are your thoughts?

ekraffmiller:
> > @GermanSaracca There is still a difference between the form validation and the API validation - from my testing, I think the API validation is requiring that the email address be from a list of top level domains, for example: https://data.iana.org/TLD/tlds-alpha-by-domain.txt.
> > When I try an address that is not in the TLD list, I get an error: Screenshot 2024-05-23 at 9 23 46 PM The API uses an apache commons validator: https://github.com/IQSS/dataverse/blob/develop/src/main/java/edu/harvard/iq/dataverse/validation/EMailValidator.java#L13 I think it is using this underlying domain validator: https://commons.apache.org/proper/commons-validator/apidocs/org/apache/commons/validator/routines/DomainValidator.html
> > I did some searches and couldn't find a javascript library that does something similar, so I'm not sure the best way to handle this, what do you think?
> 
> Hi @ekraffmiller , your findings are totally correct, there is no Javascript library that does the same validations as the Apache library ( at least not a maintained one ). One option, would be to hardcode all possible TLDS in the SPA but I really don like it because they are always changing or new ones are added. I think this is just a client side validation layer and after receiving the error we are showing in the alert so I think it is enough for the user to know what is happening. What are your thoughts?



> > @GermanSaracca There is still a difference between the form validation and the API validation - from my testing, I think the API validation is requiring that the email address be from a list of top level domains, for example: https://data.iana.org/TLD/tlds-alpha-by-domain.txt.
> > When I try an address that is not in the TLD list, I get an error: Screenshot 2024-05-23 at 9 23 46 PM The API uses an apache commons validator: https://github.com/IQSS/dataverse/blob/develop/src/main/java/edu/harvard/iq/dataverse/validation/EMailValidator.java#L13 I think it is using this underlying domain validator: https://commons.apache.org/proper/commons-validator/apidocs/org/apache/commons/validator/routines/DomainValidator.html
> > I did some searches and couldn't find a javascript library that does something similar, so I'm not sure the best way to handle this, what do you think?
> 
> Hi @ekraffmiller , your findings are totally correct, there is no Javascript library that does the same validations as the Apache library ( at least not a maintained one ). One option, would be to hardcode all possible TLDS in the SPA but I really don like it because they are always changing or new ones are added. I think this is just a client side validation layer and after receiving the error we are showing in the alert so I think it is enough for the user to know what is happening. What are your thoughts?



> I think this is just a client side validation layer and after receiving the error we are showing in the alert so I think it is enough for the user to know what is happening. What are your thoughts?

Yes, I think so too, it would be very cumbersome to maintain a list in our code.  The new alert messages work very well to inform the user 👍

ekraffmiller:
@GermanSaracca  There are some warnings due to the new lint tests...I know we don't need to fix them all, but maybe this PR could fix the ones related to Create Dataset?
```
/Users/ellenkraffmiller/Documents/GitHub/dataverse-frontend2024/src/sections/create-dataset/CreateDataset.tsx
  44:6  warning  React Hook useEffect has a missing dependency: 'setIsLoading'. Either include it or remove the dependency array  react-hooks/exhaustive-deps

/Users/ellenkraffmiller/Documents/GitHub/dataverse-frontend2024/src/sections/create-dataset/MetadataBlockFormFields/MetadataFormField/Fields/PrimitiveMultiple.tsx
  51:5  warning  React Hook useMemo has a missing dependency: 'builtFieldNameWithIndex'. Either include it or remove the dependency array  react-hooks/exhaustive-deps
```

GermanSaracca:
I completely agree, I will fix those eslint warnings, thanks @ekraffmiller

ekraffmiller (approved):
LGTM!

GPortas (approved):
LGTM

https://github.com/IQSS/dataverse-frontend/assets/17010447/ab384df4-32aa-4aa2-8011-575a2181c217

PR #409, opened 2024-05-23
Fix - Collection page datasets not being refetched after navigating from a sub collection.

### LINKED ISSUE #408: Bug - Collection page datasets not being refetched after navigating from a sub collection.
In the video you can see the unexpected behavior, when navigating to a collection other than root and navigating back to the root collection, the new datasets are not loaded.
[Dataverse.webm](https://github.com/IQSS/dataverse-frontend/assets/64560524/516d5b16-1e97-4375-b1f1-48fdabc6b464)

SPA users

GermanSaracca:
Fix bug about datasets from collection not changing when collection changed.

Closes #408 

Navigate to a dataset in a sub-collection, then click on the breadcrumbs to go to the parent collection for that dataset.
The datasets in that collection will be displayed, and then click on the breadcrumb root to display the new datasets in the root collection.

No.

No.

**Additional documentation**:
No.

### DIFF
--- a/src/sections/collection/Collection.tsx
+++ b/src/sections/collection/Collection.tsx
@@ -52,9 +52,18 @@ export function Collection({
           </>
         )}
         {infiniteScrollEnabled ? (
-          <DatasetsListWithInfiniteScroll datasetRepository={datasetRepository} collectionId={id} />
+          <DatasetsListWithInfiniteScroll
+            datasetRepository={datasetRepository}
+            collectionId={id}
+            key={id}
+          />
         ) : (
-          <DatasetsList datasetRepository={datasetRepository} page={page} collectionId={id} />
+          <DatasetsList
+            datasetRepository={datasetRepository}
+            page={page}
+            collectionId={id}
+            key={id}
+          />
         )}
       </Col>
     </Row>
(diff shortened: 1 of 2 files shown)


ekraffmiller (changes requested):
There is one difference in the Chromatic build,    https://www.chromatic.com/test?appId=646f68aa9beb01b35c599acd&id=664f41fbb2649c10c264b47b which I think is caused by Chromatic taking the snapshot before the page finishes loading. To avoid that other stories have a parameter to delay the snapshot, for example, here:
https://github.com/IQSS/dataverse-frontend/blob/2ec048c2098f8d01c3485e3bb4a8df9a8134df08/src/stories/dataset/Dataset.stories.tsx#L24C3-L27C4
I think if we add that delay to CreateDataset.stories.tsx, the Chromatic build will pass

GermanSaracca:
@ekraffmiller I added the delay (_also rename story file from DatasetCreate to CreateDataset_), I see now that snapshots are correctly being captured after the delay.
In the case of the Default Story, the snapshots shows now the actual form and not only the loading skeletons. Thanks!

ekraffmiller (approved):
Looks good!  I accepted the Chromatic changes.  Approved!

GPortas (approved):
Works good!

https://github.com/IQSS/dataverse-frontend/assets/17010447/82644473-95cc-4f27-bd5b-e69c2064efe4

Issue #363, opened 2024-04-08
Integrate the js-dataverse not found error


GPortas:
## Overview of the Feature Request

Identify the new NotFoundError in the catch block of the use case calls and handle it accordingly.

## What kind of user is the feature intended for?

This feature is intended for all users.

## What inspired the request?
- https://github.com/IQSS/dataverse-client-javascript/issues/142
- https://github.com/IQSS/dataverse-frontend/pull/352#discussion_r1549779417

## What existing behavior do you want changed?

None

## Any brand new behavior do you want to add to Dataverse?
Integrate the js-dataverse not found error

## Any open or closed issues related to this feature request?

- https://github.com/IQSS/dataverse-client-javascript/issues/142

PR #442, opened 2024-07-23
321 default dataset version

### LINKED ISSUE #321: Fix retrieving the default version from the dataset page, which is now :latest

Instead of returning :latest by default, we should replicate the behavior in JSF.

In JSF if you have permissions to view :draft but a published version exists, the published version should be opened instead of :draft, so :latest does not work.

Review other scenarios to ensure we fully replicate JSF behavior

SPA user

## What inspired the request?
- Frontend weekly meeting

Default dataset page version when no version is specified in the URL

ekraffmiller:
Changes the default dataset version requested from :latest to :latest-published.  This is needed to match the behavior of the JSF.  In the JSF, If a dataset has a draft version and a published version,  View Dataset page shows the published version by default (when there isn't a version search param in the url).

Closes #321 

Now that the default version is :latest-published, in order to view the draft version of a dataset, you need to add 'version=DRAFT' to the view dataset url.  (The draft version used to be displayed by default when the user has permission to view it).

Create a dataset that has a published version and a draft version.  Test that the version displayed in the View Dataset page is the published version, when no version search param is used.

no

**Additional documentation**:

### DIFF
--- a/src/dataset/domain/models/Dataset.ts
+++ b/src/dataset/domain/models/Dataset.ts
@@ -214,7 +214,7 @@ export enum DatasetNonNumericVersion {
   DRAFT = ':draft',
   LATEST_PUBLISHED = ':latest-published'
 }
-export enum DatasetVersionNonNumericSearchParam {
+export enum DatasetNonNumericVersionSearchParam {
   DRAFT = 'DRAFT'
 }
 
--- a/src/dataset/domain/models/Dataset.ts
+++ b/src/dataset/domain/models/Dataset.ts
@@ -211,7 +211,11 @@ export enum DatasetPublishingStatus {
 
 export enum DatasetNonNumericVersion {
   LATEST = ':latest',
-  DRAFT = ':draft'
+  DRAFT = ':draft',
+  LATEST_PUBLISHED = ':latest-published'
+}
+export enum DatasetVersionNonNumericSearchParam {
+  DRAFT = 'DRAFT'
 }
 
 export class DatasetVersionNumber {
--- a/src/dataset/infrastructure/repositories/DatasetJSDataverseRepository.ts
+++ b/src/dataset/infrastructure/repositories/DatasetJSDataverseRepository.ts
@@ -1,5 +1,5 @@
 import { DatasetRepository } from '../../domain/repositories/DatasetRepository'
-import { Dataset } from '../../domain/models/Dataset'
+import { Dataset, DatasetNonNumericVersion } from '../../domain/models/Dataset'
 import {
   getDataset,
   getAllDatasetPreviews,
@@ -116,9 +116,10 @@ export class DatasetJSDataverseRepository implements DatasetRepository {
 
   getByPersistentId(
     persistentId: string,
-    version?: string,
+    version: string = DatasetNonNumericVersion.LATEST_PUBLISHED,
     requestedVersion?: string
   ): Promise<Dataset | undefined> {
+    console.log('getByPersistentId', persistentId, version, requestedVersion)
     return getDataset
       .execute(persistentId, version, includeDeaccessioned)
       .then((jsDataset) => this.fetchDatasetDetails(jsDataset, version))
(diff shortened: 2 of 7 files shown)


src/dataset/infrastructure/repositories/DatasetJSDataverseRepository.ts
```diff
@@ -147,10 +148,14 @@ export class DatasetJSDataverseRepository implements DatasetRepository {
       })
       .catch((error: ReadError) => {
         console.error(error)
-        if (!version) {
+        if (version === DatasetNonNumericVersion.LATEST_PUBLISHED) {
           throw new Error(`Failed to get dataset by persistent ID: ${error.message}`)
         }
```
GermanSaracca, line 153: question: why only throwing the error if the requested version was :latest-published ?
I understand the need of refetching and saving the original requested version but what happen if there is an error when trying to fetch a :draft or :latest version?

src/sections/shared/form/DatasetMetadataForm/useSubmitDataset.ts
```diff
@@ -56,11 +57,12 @@ export function useSubmitDataset(
     )
 
     if (mode === 'create') {
+      const DRAFT_PARAM = DatasetVersionNonNumericSearchParam.DRAFT
       createDataset(datasetRepository, formattedFormValues, collectionId)
         .then(({ persistentId }) => {
           setSubmitError(null)
           setSubmissionStatus(SubmissionStatus.SubmitComplete)
-          navigate(`${Route.DATASETS}?persistentId=${persistentId}`, {
+          navigate(`${Route.DATASETS}?persistentId=${persistentId}&version=${DRAFT_PARAM}`, {
```
GermanSaracca, line 65: Just a suggestion, if DRAFT_PARAM is only going to be used in one place I think we don't need to add it to a constant variable, just using the enum directly in the url seems fine to me. e.g: 
```
navigate(
  `${Route.DATASETS}?persistentId=${persistentId}&version=${DatasetVersionNonNumericSearchParam.DRAFT}`,
  {
    state: { created: true },
  }
)
```

GermanSaracca (commented):
Looks good! Just a question and some suggestions only.

src/dataset/infrastructure/repositories/DatasetJSDataverseRepository.ts
ekraffmiller: The assumption with this logic is that the error was thrown because the version doesn't exist.  We don't currently have a specific error type to catch when the version isn't found, but there are plans to add one: https://github.com/IQSS/dataverse-frontend/issues/363 .  You're right that we should have separate logic for handling other types of errors, but I thought it was better to keep it like this for now, rather than adding logic that will be replaced later.  So what will happen for now, is that the UI will show 'page not found' if both the requested version and the :latest-published version throw an error, for whatever reason.  Let me know what you think, if that's ok, or we should add more logic for handling different types of errors.

src/dataset/domain/models/Dataset.ts
```diff
@@ -211,7 +211,11 @@ export enum DatasetPublishingStatus {
 
 export enum DatasetNonNumericVersion {
   LATEST = ':latest',
-  DRAFT = ':draft'
+  DRAFT = ':draft',
+  LATEST_PUBLISHED = ':latest-published'
+}
+export enum DatasetVersionNonNumericSearchParam {
```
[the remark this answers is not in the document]
ekraffmiller: Yes, makes sense!

ekraffmiller:
Thanks @GermanSaracca I made the suggested changes - the component test is failing in the last run only because I had to re-run it, and the coveralls script failed the second time since the coverage report had been created on the first run.

I'm not sure why the component tests had to be re-run, it seems like maybe a failure to get the stock images for the metadata test?

src/dataset/infrastructure/repositories/DatasetJSDataverseRepository.ts
GermanSaracca: You are right, I think it is fine to leave it as it is for now, as you mention this will probably change with the 363 integration.

GermanSaracca:
> I'm not sure why the component tests had to be re-run, it seems like maybe a failure to get the stock images for the metadata test?

What do you mean by stock images ? I couldn't find that part of the code, are we using an external images library in some test?

Approving 🚀 , we can discuss about those test for another iteration I believe 👍🏼

GPortas (approved):
LGTM!

https://github.com/user-attachments/assets/054049f6-a6f9-4df2-9d2d-e964ec7b8071

PR #459, opened 2024-08-13
feat: CollectionCard for Collection Page

### LINKED ISSUE #455: Add Collection Card for Collections Page
This component is for Collections (Dataverses in JSF) that appear in the search results of the collection page: 
Screenshot 2024-07-29 at 11 23 37 AM

ekraffmiller:
Card Component to display Collections in Collection Page search results 

Closes #455 

The code for the 'Unpublished' label uses DatasetLabels, but it isn't related to Datasets.  I was thinking the label code could be refactored to be more generic, but since that's a big change, I left it as is for now.

`npm run storybook` and view stories here: http://localhost:6006/?path=/story/sections-collection-page-collectioncard--default

**Additional documentation**:


g-saracca (changes requested):
Hi @ekraffmiller , code looks good!
About `DatasetLabels` , I noticed that they are also used for `FileCard`.
If they all share the same properties, labels and semantic values it's fine to make a generic one in a shared folder maybe? but if they don't, then I would create a FileLabels and a CollectionLabels as a first step.
After some time, if we realize that we can have one for all DV objects, then we make a more generic one, but for now I think we should create one label per domain.

 I think it's better to have things organized now, before integrating with the use case.
 What do you think?

ekraffmiller:
> Hi @ekraffmiller , code looks good! About `DatasetLabels` , I noticed that they are also used for `FileCard`. If they all share the same properties, labels and semantic values it's fine to make a generic one in a shared folder maybe? but if they don't, then I would create a FileLabels and a CollectionLabels as a first step. After some time, if we realize that we can have one for all DV objects, then we make a more generic one, but for now I think we should create one label per domain.
> 
> I think it's better to have things organized now, before integrating with the use case. What do you think?

@g-saracca , I looked more at the File information, and I realized that it already has FileLabels, which contain different information from DatasetLabels, and both are needed in the FileCard.  So I updated the card to show that.  For the Collection Card label, since it is only one label with simple logic, I did the same thing you have in the Collection Page, and used the Badge component directly.  Let me know what you think. thanks!

g-saracca (approved):
Looks good!

GPortas (approved):
All storybook cases are correctly replicating JSF.

Merging.

PR #464, opened 2024-08-22
Feature/417 File upload integration

### LINKED ISSUE #417: Integrate file upload page with js-dataverse use cases

Integrate the file upload page UI from [GitHub issue 366](https://github.com/IQSS/dataverse-frontend/issues/366) with the js-dataverse use cases implemented on https://github.com/IQSS/dataverse-client-javascript/pull/153.

We don't necessarily have to wait for https://github.com/IQSS/dataverse-client-javascript/pull/153 to be merged for working on the integration since we already have an npm version generated from the PR.

SPA user.

UI and js-dataverse work.

None

Functional file upload feature

- https://github.com/IQSS/dataverse-frontend/issues/366
- https://github.com/IQSS/dataverse-client-javascript/issues/152

ErykKul:
File upload works now, integration test is added

- Closes #417 

Upload some file to a dataset.

No interface changes.

Something like "added files upload feature"


src/sections/upload-dataset-files/UploadDatasetFiles.tsx
```diff
@@ -42,15 +46,36 @@ export const UploadDatasetFiles = ({ fileRepository: fileRepository }: UploadDat
...
     setUploadingToCancelMap((x) => {
-      x.delete(key)
+      x.delete(FileUploadTools.key(file))
       return x
     })
     releaseSemaphore(file)
   }
 
+  const fileUploadFinished = (file: File) => {
+    const hash = md5.create()
```
g-saracca, line 58: Just to understand the code better, why do we need to store the checksum of the uploaded file?
And notice that in the .finally block, the code may succeed or may fail, the finally block runs when the promise is settled but runs on both succeed or fail of the promise

src/files/domain/useCases/addUploadedFiles.ts
```diff
@@ -14,18 +14,3 @@ export function addUploadedFiles(
     })
     .finally(done)
```
g-saracca, line 15: The `done` function is being called in the `.finally` block here, but that is not a guarantee that the use case ran correctly,  the finally block just tells you that the promise is settled regardless of whether it was successful or not.
Instead, try with this:

```
  fileRepository
    .addUploadedFiles(datasetId, files)
    .then(done)
    .catch((error: Error) => {
      throw new Error(error.message)
    })
```

g-saracca (changes requested):
Hi @ErykKul great work! just some comments.
Also one thing I noticed in the UI, the checkboxes to select an uploaded file from the UploadedFiles list are not being checked/unchecked on the first click, somehow they only are checked/unchecked on the second click.

src/sections/upload-dataset-files/UploadDatasetFiles.tsx
ErykKul: Uploading files to a dataset is a two-step process: first upload all files to s3, then add all uploaded files (this call requires the checksum) to the dataset. The checksum is needed when we call "add uploaded files" use-case, but it is calculated immediately after uploading a specific file to s3 ("upload file" use-case), while the semaphore is still taken. The semaphore makes sure that we can upload/calculate checksum only for a certain number of files concurrently.

That said, reading a file should not fail while calculating a checksum, unless you delete that file after it was uploaded to s3 on your computer, but before the checksum calculation finished. In that case, the value for checksum is not set and addUploadedFiles will fail. Regardless of the success of checksum calculations, we need to prevent the resource leak (by keeping the reference to the cancel function that can no longer be used, as the upload to s3 finished) and keeping the semaphore (that would freeze the application, as the upload to s3 is queued on that semaphore). The `finally` block release these resources.

g-saracca: Thanks for the detailed explanation @ErykKul 👏🏼 👏🏼

g-saracca:
Hi @ErykKul , thanks for addressing the requested changes. 
I am seeing fails in the e2e tests, could you take a look at them?
If they keep failing try increasing the cypress command timeout like I did in [this PR](https://github.com/IQSS/dataverse-frontend/pull/466/files#diff-3f62e885350a48939f240b9daf40f6b06d41da7751f2b613238fc7669ac0a1b2) maybe that will help.

ErykKul:
@g-saracca Thanks! It fixed the test that was initially failing. I did notice that my file upload test was also failing after the DTO refactoring, I fixed it as well. It looks like all tests are passing now.

g-saracca (dismissed):
Looks good! Approving

ErykKul:
I could not reproduce the .java files upload issue, I uploaded a bunch of them, without any problem. I used Firefox, I will also try Chrome and see from there. I do not see anything specific to the file type preventing some kinds of files from being uploaded, the file type is hard-coded to `'Content-Type': 'application/octet-stream'` in [DirectUploadClient.ts](https://github.com/IQSS/dataverse-client-javascript/blob/develop/src/files/infra/clients/DirectUploadClient.ts)

As for the content length header: it is also being set in [DirectUploadClient.ts](https://github.com/IQSS/dataverse-client-javascript/blob/develop/src/files/infra/clients/DirectUploadClient.ts) (line 69). I assume that the warning is that it was not set? The upload itself works, I get files uploaded and downloaded, so I assume it is not the cause of the problems...

The progress is also set from the same file, only the calls from the lines 46 and 53 work. It looks like the progress update at line 119 never gets called. So the progress is first set to 10% when upload starts, and to 100% when its finished. It looks like the UI does exactly that.

At the first glance, the problems you mention do not seem to be related to this PR.

ErykKul:
I could reproduce the problem with chromium, with firefox I can upload java files, I can't with chromium. Strange...

ErykKul:
It turns out that FireFox does recognize java files and gives them a correct content-type, but chromium does not. I have hard-coded it to `application/octet-stream`, it fixed the upload problem.

ErykKul:
`application/octet-stream` is the default type when uploading, it basically means that we let Dataverse figure it out, either through the configuration or with JHOVE. I checked, the uploaded files get the correct mime-type after upload. I think I confused the header being passed to the s3 with the file type passed to Dataverse, the one passed to Dataverse was not hardcoded, but set to a value chosen by the browser. That is why it worked with FireFox, but not with Chromium.

ErykKul:
@GPortas Can you look at it again? Thanks!

ErykKul:
I just discovered another problem: when uploading many files, the upload can get stuck. I am looking at it now.

ErykKul:
@qqmyers  Chromium was giving an empty string as mime type, which was (I guess) rejected by Dataverse, I can try detecting that specific case - invalid mime type provided by the browser. I will also dig deeper into it and check the log of Dataverse to see what exactly happened.

ErykKul:
@g-saracca The problem I am having with stuck uploads is very similar to the checkbox problem I was having and that you noticed; it worked fine with mocks, but worked incorrectly in the real SPA (I see the same behavior for the semaphores: function being wrongly called several times, while in the code only one call is possible, the func inside the onChange was also called multiple times for the checkbox, while there was only one change). Together with the mime-type thing, I need to go deeper into it, so I am moving this PR back to "in progress"

qqmyers:
Ah - OK - we might not handle an empty/missing value, so checking and assuming that's application/octet-stream on the client or server side might be better than just always hardcoding.

g-saracca:
Hi @ErykKul please point me out the part of the semaphore code or function that it is being called multiple times maybe I can find out what is the problem.
I do think this part of the code below is a little bit hard to control with a useEffect, as it will re run everytime fileUploaderState changes, and as it's not a primitive it will run all the time because it cant really compare it to the previous state. Maybe you can find a solution that doesn't require a useEffect.
```
  useEffect(() => {
    setFiles((newFiles) =>
      newFiles.filter((x) => {
        const res = !FileUploadTools.get(x, fileUploaderState).removed
        if (!res) {
          cleanFileState(x)
        }
        return res
      })
    )
  }, [fileUploaderState, cleanFileState])
```
Try, just as an experiment, using useDeepCompareEffect (`import { useDeepCompareEffect } from 'use-deep-compare'`) instead of useEffect, it will deep compare fileUploaderState to detect a change, but checkout the performance monitor with the dev tools please because it can consume quite a lot of resources perhaps by uploading several files.

ErykKul:
@qqmyers 
I updated the code and the octet-stream is now only used if browser failed to detect the mime type (it fix the problem with chromium and .java files):
```js
uploadedFile.fileType === '' ? 'application/octet-stream' : uploadedFile.fileType
```
@g-saracca 
Thanks! I used that in the code, it does not look like it had much impact, but it pointed me to where the real problem was! It was the upload effect that run each time when `files` state was changed, that on its turn was changed on each `fileUploadState` change. I simplified the code and replaced my semaphore by a real one from `async-mutex` package. The code looks much better now and runs more smoothly. Can you re-review? Thanks!

g-saracca:
Hi @ErykKul great!
Could be nice to add a comment on this line on why the need for a ternary, so future devs knows why.
`uploadedFile.fileType === '' ? 'application/octet-stream' : uploadedFile.fileType`

If you think `useDeepCompareEffect` was not part of the solution please remove it and check if everything keeps working correctly, what do you think?

Here 👇 , `setFiles((newFiles) =>` the callback args `(newFiles)`  should be `(currentFiles)` as the callback of a stateSetter always have the current state, so is clear what this code is doing.
```
  useDeepCompareEffect(() => {
    setFiles((newFiles) =>
      newFiles.filter((x) => {
        const res = !FileUploadTools.get(x, fileUploaderState).removed
        if (!res) {
          cleanFileState(x)
        }
        return res
      })
    )
  }, [fileUploaderState, cleanFileState])
```

ErykKul:
I reverted to the `useEffect`, added the comment, etc. I also removed the test that failed. It is less useful now, since the custom semaphore is no longer used and does not need to be tested. It also turned out to be flaky, since I had rerun all the tests locally and they all passed.

ErykKul:
The integration test that failed is not linked to the file upload code. It might be related to the s3 storage? In worst case, after each integration test run we leave some garbage on the s3 bucket. But even if that is the case, I am still not sure why that test suddenly failed. Maybe I should increase the timeout for the integration tests again?

g-saracca:
Hi @ErykKul, things I noticed:
- Still getting a `Refused to set unsafe header "Content-Length"` error in console when uploading a .jpg file.
- Progress bar doesn't seem to we showing the progress updating correctly, just increase one time and then the file is completely uploaded and removed from the file drop zone.

ErykKul:
I have added the gap. I think that the header warning is something that the js-client sets. I can look at it, as well as improving the progress updates (I think that one would also require changes at the js-client).

GPortas:
@ErykKul Thanks for addressing the changes. I have retested the feature for .java files and it works great.

I think this is ready to be merged, and open a separate issue in js-dataverse to review the header warning and progress updates.

GPortas (dismissed):

ErykKul:
I tried fixing the header and progress bar issue: https://github.com/IQSS/dataverse-client-javascript/pull/187
I will test it locally now. I will let you know if it worked. It is fine for me if it will be a separate issue.

ErykKul:
I think it fixed it, I will add it to this PR, we can revert it if needed.

ErykKul:
image

The progress bar moves nicely and there are no errors in the log. However, multipart upload does not work as it should, it looks like it "resets" with every part (starts from 0 and never reaches 100%). I will try fixing it, it should be easy to do (I think that my assumption on the index and part size were incorrect, or I messed up the formula).

g-saracca (dismissed):

ErykKul:
I think that it was a correct behavior: resetting was due to the upload failing, from the log it looks like uploading first part always failed with some retries (maybe the file is too large for the test-bucket)? Anyway, it is the best I could do for now. I updated the dependency again, after a small fix on the multipart progress updates. I will stop committing now, such that the review and QA can continue. Thanks @g-saracca and @GPortas

g-saracca:
@ErykKul you made changes to the file upload use case on the js-dataverse repo?

g-saracca:
If that is the case we should first I think , review and approve the JS-dataverse PR 👍🏼

ErykKul:
Yes, it was a small change in the js-dataverse. It is not urgent, I think it is most important to get it right, so I think it is a good idea to take a look at my change at the js-dataverse first: [PR 187](https://github.com/IQSS/dataverse-client-javascript/pull/187)

GPortas (approved):
This works great now. I just merged the js-dataverse PR and tested it on this branch. All issues are solved.

https://github.com/user-attachments/assets/520f5997-c321-46df-8e90-f0daf9979c2f

PR #477, opened 2024-09-02
New homepage routing + search input

### LINKED ISSUE #476: New Homepage Routing + Search Input
Create the initial UI work for the new homepage for the SPA.
This includes:

- Routing configuration and link changes on other parts of the app.
- Custom search input, submitting the search (by pressing the enter key or clicking the button) will perform a navigation to the root collection page with the encoded search query placed in the url as query parameter `?q=[value]`.

All SPA users.

Q3 Roadmap

Show this new homepage as main entry instead of showing the root collection page.

[Search homepage design](https://github.com/IQSS/dataverse-frontend/issues/472)

g-saracca:
Initial UI work for the new homepage for the SPA.
This includes:
- Routing configuration and link changes on other parts of the app.
- Custom search input, submitting the search (by pressing the enter key or clicking the button) will perform a navigation to the root collection page with the encoded search query placed in the url as query parameter ?q=[value].

- Closes #476 

This PR only adds the search input from the new home page. Submitting the search (by pressing the enter key or clicking the button) will perform a navigation to the root collection page with the encoded search query placed in the url as query parameter but won't perform a call to any use case with that search value.

I have also add a box shadow to the navbar as it looks in the JSF version, to make it easier to distinguish the header from what is going on underneath.
I have changed the position of the navbar from fixed to sticky, this way it occupies a height and it is not necessary to give a margin top to the main container so that it is not hidden underneath initially _(this doesn't create a visual change)_.
I added a padding top of 8px to the breadcrumbs so they are not so close to the navbar.
I refactor some e2e tests because now '/' route is now the homepage and not the root collection.
First item of breadcrumbs now goes to the collections base route '/collections' and not to '/' as it is the homepage.

1. Execute `npm i`.
3. Return with `cd ../../`.

**Step 2: Test the search input**

- Submit different entries in the search and confirm that the value is correctly added as a query param in the url of the collections page, the search value is encoded, so to actually test that the value is correctly saved on the url you could copy the value after the `?q=` and decoded in the browser dev console by pasting following code and checking the value of `decodedQuery`.
```
const searchParams = new URLSearchParams(location.search)
const query = searchParams.get('q')
const decodedQuery = decodeURIComponent(query)
```
## Does this PR introduce a user interface change? If mockups are available, please link/include them here: Yes.
Screenshot 2024-09-02 at 15 53 44

No
## Additional documentation:
No


packages/design-system/src/lib/components/form/form-group/form-element/FormInput.tsx
```diff
@@ -26,6 +27,7 @@ export const FormInput = React.forwardRef(function FormInput(
     value,
     required,
     autoFocus,
+    autoComplete,
```
ekraffmiller, line 30: Just curious why autoComplete was added here? I see that it is a property in SearchInput.tsx, but it is set to "off".

src/sections/homepage/search-input/SearchInput.tsx
```diff
@@ -0,0 +1,68 @@
...
+          aria-label="Search"
+          autoFocus
+          autoComplete="off"
+          className={styles['text-input']}
+          value={searchValue}
+          onChange={handleSearchChange}
+          ref={inputSearchRef}
+        />
+        {searchValue && (
+          <CloseButton
```
ekraffmiller, line 55: Can we use the design system button here rather than a react-bootstrap component?

src/sections/homepage/search-input/SearchInput.tsx
```diff
@@ -0,0 +1,68 @@
+import { useRef, useState } from 'react'
+import { useNavigate } from 'react-router-dom'
+import { Form } from '@iqss/dataverse-design-system'
+import { CloseButton } from 'react-bootstrap'
+import { Search as SearchIcon } from 'react-bootstrap-icons'
```
ekraffmiller, line 5: This may be a lower level priority, but I think our goal is to avoid referring to react-bootstrap in the SPA code, and keep it in the design system code, is that right?
After seeing this I noticed that we are importing 'react-bootstrap-icons' in several places, and one instance where we are importing Dropdown from react-bootstrap.

ekraffmiller (changes requested):
The page looks nice! I have some comments about how we are using react-bootstrap.  Also, need to add Stories and component tests.

packages/design-system/src/lib/components/form/form-group/form-element/FormInput.tsx
g-saracca: Yes, this is because the current `FormInputProps` interface wasn't allowing me to pass the native attribute autoComplete, so I explicitly declared it.
I am using it in the search input with an "off" value to avoid the browser to provide automated assistance in filling out the search input.
I think it's a better experience without it.

src/sections/homepage/search-input/SearchInput.tsx
g-saracca: Yeah, I will add the CloseButton to the Design System, it looks better that using our Button with a close icon.
About the Dropdown used in the AddData, I didn't know about that, I will address it better at another time.

g-saracca:
@ekraffmiller all changes have been addressed, story, tests and close button to the design system, thanks for the catch!

ekraffmiller (approved):
looks good, approved!

GPortas (approved):
Looks great! A good starting point for the search home page.

PR #478, opened 2024-09-04
File upload: alert when a file could not be added for upload

### LINKED ISSUE #467: Feature Request/Idea: better handling of duplicate files when adding files to upload
Current behavior:
- adding file with the same name (and label) as a file that is already being added results in the new file being silently ignored
- files with different names but with the same checksum are being added without a warning

Possible improvements
- error is given (e.g., popup) when a file with the same name is already being added
- warning is given when a file with the same checksum is already being added

It might be also useful to extend the checks for identical files to the files already present in the dataset. It is currently not done. For example, when a file with a name and label already exists in the dataset, the new file gets a suffix "-1" (e.g., "test-1.json" when "test.json" is already in the dataset). A warning could be given before adding such file. Also, a warning should be given if a file with the same checksum is already present in the dataset. However, this might slow down uploads for large datasets containing thousands of files, and when uploading only few new files. Therefore, this remains out of scope for this issue.

Depositor

File upload feature as it is implemented now.

Better edge case handling when uploading files.

File upload is a new feature that is currently being developed.

#431

ErykKul:
With each upload, all files need to have a unique combination of the label (directory) and the file name. Until now, when adding the same file (as already present in the list of files to be added at the bottom of the file upload page), it is silently ignored and is not uploaded. This PR adds an alert on the top of the page when this happens to notify the user that the file could not be added.

Note that you can still add the same file multiple times to the dataset by uploading them one by one. The upload page does not check the file names of the files already present in the dataset. When the same file is uploaded again, it is handled in the standard way by the Dataverse API and gets the "-1" suffix, e.g., after adding "test.json" twice, you will have two files in the dataset: "test.json" and "test-1.json".

- Closes #467

Add the same file for upload twice, when adding second time (and the previously added files is still in the list of the files to be added at the bottom of the upload page), an alert should be shown (and the file is not added second time).

### DIFF
--- a/public/locales/en/uploadDatasetFiles.json
+++ b/public/locales/en/uploadDatasetFiles.json
@@ -60,5 +60,8 @@
     "selectTags": "Select tags to add",
     "saveChanges": "Save Changes",
     "cancelChanges": "Cancel Changes"
+  },
+  "alerts": {
+    "fileNameNotUnique": "File {{name}} was already added and is ignored"
   }
 }
--- a/src/sections/upload-dataset-files/FileUploader.tsx
+++ b/src/sections/upload-dataset-files/FileUploader.tsx
@@ -4,6 +4,8 @@ import { ChangeEventHandler, DragEventHandler, useEffect, useRef, useState } fro
 import { Plus, X } from 'react-bootstrap-icons'
 import { FileUploadTools, FileUploaderState } from '../../files/domain/models/FileUploadState'
 import styles from './FileUploader.module.scss'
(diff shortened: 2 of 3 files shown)


ErykKul:
After some feedback, this one needs more work before it is ready for review -> status changed to in progress.

cmbz:
2024/09/25: Moving back to SPA backlog for further work.

g-saracca:
Closing PR, this scenario is already handled by the file uploader.

PR #479, opened 2024-09-05
Feat/470 user panel api token

### LINKED ISSUE #470: User Panel UI + API Token Tab content
This is the initial work for the user panel in the SPA.
This work will include:

- Add links to the user panel in the User dropdown from the application header.
- Add routing to the user panel.
- Create 4 tabs panel with tab open by default according to url query param.
- Only the API Token tab content will be created, the rest of the tabs will remain empty for now, maybe disabling tabs.

Note that this is only UI work, it wont show real info about API token yet.

All logged in users.

Q3 Roadmap Nices To Have

g-saracca:

> This is the initial work for the user panel in the SPA.
> This work will include:
> 
> Add links to the user panel in the User dropdown from the application header.
> Add routing to the user panel.
> Create 4 tabs panel with tab open by default according to url query param.
> Only the API Token tab content will be created, the rest of the tabs will remain empty for now, maybe disabling tabs.
> Note that this is only UI work, it wont show real info about API token yet.

This part of the documentation [Users Token Management](https://guides.dataverse.org/en/latest/api/native-api.html#users-token-management) could be useful for developing use cases for managing api tokens from the SPA.

- Closes #470 

The rest of the tabs are disabled, only the API Token tab is enabled.
Also the Recreate Token and Revoke Token buttons are disabled.

1. Execute `npm i`.
3. Return with `cd ../../`.

**Step 2: Test the feature**
This is only UI work, but couple things could be tested.

- Check that you navigate successfully to the Account page with the apiToken tab selected by default when clicking on the Api Token item on the user dropdown in the header.
- Check that the fake api token is correctly being copied to the clipboard

Screenshot 2024-09-05 at 13 19 05


public/locales/en/account.json
```diff
@@ -0,0 +1,18 @@
+{
+  "pageTitle": "Account",
+  "tabs": {
+    "myData": "My data",
+    "notifications": "Notifications",
+    "accountInformation": "Account information",
+    "apiToken": "API token"
```
pdurbin, line 7:
```suggestion
    "myData": "My Data",
    "notifications": "Notifications",
    "accountInformation": "Account Information",
    "apiToken": "API Token"
```

Do we want Title Case or not? 🤔 Do we have any docs on which to use? I checked the Style Guide but I didn't see anything. In JSF we used Title Case for these:

Screenshot 2024-09-06 at 9 26 18 AM

pdurbin (commented):
Quick question about text style.

public/locales/en/account.json
g-saracca: Hi @pdurbin, yes indeed !! I will commit your suggestion, many thanks!

src/sections/account/Account.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+const tabsKeys = AccountHelper.ACCOUNT_PANEL_TABS_KEYS
+
+export const Account = () => {
+  const { t } = useTranslation('account')
+  const { setIsLoading } = useLoading()
+  const [searchParams] = useSearchParams()
+  const defaultActiveTabKey = AccountHelper.defineSelectedTabKey(searchParams)
+
+  const rootHierarchy = UpwardHierarchyNodeMother.createCollection({
```
ekraffmiller, line 20: The name and alias of the root collection is not guaranteed to be 'root', so I think we need to replace this with a call to get the root collection from the API.  There is a special keyword, ':root', that returns the root collection: https://guides.dataverse.org/en/latest/api/native-api.html#view-a-dataverse-collection.  So we should also change the default ROOT_COLLECTION_ALIAS to ':root'.

src/sections/account/api-token-section/ApiTokenSection.tsx
```diff
@@ -0,0 +1,71 @@
...
+                rel="noreferrer"
+              />
+            )
+          }}
+        />
+      </p>
+      {apiToken ? (
+        <>
+          <p className={styles['exp-date']}>
+            {t('expirationDate')} <time dateTime={expirationDate}>{expirationDate}</time>
```
ekraffmiller, line 40: If the expiration date is expired, we need a warning message about that, for example:

Screenshot 2024-09-12 at 10 19 19 AM
Screenshot 2024-09-12 at 10 19 19 AM

src/sections/account/Account.tsx
```diff
@@ -0,0 +1,58 @@
...
+  useEffect(() => {
+    setIsLoading(false)
+  }, [setIsLoading])
+
+  return (
+    <section>
+      <BreadcrumbsGenerator hierarchy={rootHierarchy} withActionItem actionItemText="Account" />
+
+      <header>
+        <h1>{t('pageTitle')}</h1>
```
ekraffmiller, line 34: The page title in JSF includes the name of the root collection, but I don't think we need it, because of the breadcrumbs. What do you think?
Screenshot 2024-09-12 at 10 19 10 AM

ekraffmiller (changes requested):
Looks good! just need small updates and also a Story

src/sections/account/Account.tsx
g-saracca: hmm thats true, interesting, I think the change will require a completely separate issue (maybe a js-dataverse use case first), as it will be quite large, I think.
We rely on a Root word and root alias in several places of the application at this point.

src/sections/account/api-token-section/ApiTokenSection.tsx
g-saracca: Yes, that will require also a use case from this native API endpoint [Find a Token’s Expiration Date](https://guides.dataverse.org/en/latest/api/native-api.html#id162).
I prefer to add that message when we make the integration, but I will mention about this use case in [this current sprint issue](https://github.com/IQSS/dataverse-client-javascript/issues/190).

src/sections/account/Account.tsx
g-saracca: Now thinking... we will also need to use that "dynamic" root alias in the URL paths and not the hardcoded root, quite a challenge for an SPA, so we really need to first create a use case for it and then make all the necessary changes in the SPA, maybe not a priority right now? but for sure an really important thing.

src/sections/account/Account.tsx
g-saracca: Yes, it seemed a bit repetitive, that's why I didn't add it, I think we can leave it for now.

g-saracca:
I added a basic story, when we integrate with the use cases we will be able to mock responses to show different scenarios.
I think is ready for review now @ekraffmiller

src/sections/account/Account.tsx
ekraffmiller: Yes, you're right I think it makes sense to add another issue for that.

src/sections/account/Account.tsx
ekraffmiller: Yes, I agree! just wanted confirm

src/sections/account/Account.tsx
ekraffmiller: I added an issue for that, here: https://github.com/IQSS/dataverse-frontend/issues/489

ekraffmiller (approved):
looks good, approved!

ChengShi-1:
Hi German! I noticed that even without logging in, I was able to access the API token page by directly navigating to 'http://localhost:8000/spa/account?tab=apiToken'. Should we implement some measures to prevent unauthorized access to this page?

g-saracca:
> I noticed that even without logging in, I was able to access the API token page by directly navigating to 'http://localhost:8000/spa/account?tab=apiToken'. Should we implement some measures to prevent unauthorized access to this page?

Hi @ChengShi-1, yes, that is going to be solved when me merge the [Private Routes Guards PR](https://github.com/IQSS/dataverse-frontend/pull/483)

PR #512, opened 2024-09-30
Feature/381 implement lazy loading 2

### LINKED ISSUE #381: Implement lazy loading
It would be very beneficial to improve the performance of the application to load the different pages in a lazy way.
This could just be done at the Router level with [lazy imports](https://react.dev/reference/react/lazy)

This way, if the user only visits the Collections page, for instance, only the javascript needed for that view will be loaded and not the javascript of the whole application.

### Update:
After investigating why the Suspense fallback (spinner) component for the dynamically loaded modules was rendering more than once, it was discovered that it was due to the “useTranslation” hook that was triggering the Suspense when loading a locale json after the page was loaded. 
See [documentation info](https://react.i18next.com/latest/usetranslation-hook).
Screen Shot 2024-09-25 at 10 30 41

This is currently happening on Beta environment so this will fix this also.

g-saracca:
Applies **_react lazy imports_** for page modules. Each page within the SPA build will be a separate module with its corresponding javascript and css. This will reduce the initial load time, loading only the assets needed for the SPA requesting path.

- Closes #381 

Code visual inspection and re-run of tests actions.


tests/component/sections/collection/Collection.spec.tsx
```diff
@@ -21,6 +21,11 @@ describe('Collection page', () => {
   })
 
   it('renders skeleton while loading', () => {
+    const DELAYED_TIME = 200
```
ekraffmiller, line 24: Just curious if the change in this skeleton test is related to the lazy loading?  Or just to make sure that the timing of the test is more deterministic?

src/sections/shared/layout/app-loader/AppLoader.tsx
```diff
@@ -1,9 +1,17 @@
+import cn from 'classnames'
```
ekraffmiller, line 1: Do we need a test for this class, for the appearance of the spinner?

ekraffmiller (commented):
It looks great, and it's much faster! 🚀
Just had questions about the testing

tests/component/sections/collection/Collection.spec.tsx
g-saracca: I've always had the feeling that these tests of checking loading skeletons were a bit flaky tests.
This makes the test more robust, I found this solution to delay a stub while working on the collection page and felt it was appropriate to implement them in these tests.

src/sections/shared/layout/app-loader/AppLoader.tsx
g-saracca: Yes you are right, I will add a component test and validate that the dynamic classes based on the prop actually change the css of the container.
Also I need to solve some merging conflicts.

g-saracca:
@ekraffmiller I added a component test for the AppLoader component and solve merge conflicts 👍🏼

ofahimIQSS:
Ran tests in local, no issues discovered. This will be tested more over time.

PR #540, opened 2024-11-01
Fix/528 Use fields from parent - Create Collection form behavior

### LINKED ISSUE #539: Create Collection - if "Use metadata fields from [Parent]" is checked, send blocknames & input levels as undefined
As discussed with @pdurbin and @GPortas, we should not send metadata block names & input levels from the create collection form if "Use metadata fields from [Parent]" is checked.
Instead we should send those two properties as undefined. Before we were sending the metadata blocks and input levels that were configured in the parent collection.

g-saracca:

In the Create Collection form, when the user checks the “Use fields from [parent]” checkbox, we are now sending metadata block names and input levels as undefined to the js-dataverse use case.
This was tested also using the GHCR container image generated from https://github.com/IQSS/dataverse/pull/10985 PR and it works great.

- Closes #539 

No

No

## Additional documentation:
No

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "0.1.0",
       "dependencies": {
         "@faker-js/faker": "7.6.0",
-        "@iqss/dataverse-client-javascript": "2.0.0-alpha.2",
+        "@iqss/dataverse-client-javascript": "2.0.0-alpha.4",
         "@iqss/dataverse-design-system": "*",
         "@istanbuljs/nyc-config-typescript": "1.0.2",
         "@tanstack/react-table": "8.9.2",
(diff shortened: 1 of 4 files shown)


ekraffmiller:
This change is causing an error in Create Dataset, when the parent collection is not the Root.  We need a fix in the Dataverse API to get the correct metadata fields for the Create Dataset form.  Added a waiting tag to this issue.
Here is the error:

https://github.com/user-attachments/assets/6552d8ab-d301-4f05-9949-6fcdd1e78380

g-saracca:
@ekraffmiller working great on my end after pulling new unstable dataverse image.

https://github.com/user-attachments/assets/3cc1ffa9-56bf-4b53-995b-e9ad6e7b913f

ekraffmiller:
> @ekraffmiller working great on my end after pulling new unstable dataverse image.
> 
>  Screen.Recording.2024-11-05.at.14.49.43.mov

Working for me now too, needed to get the newest image

ekraffmiller (approved):
looks good, approving 👍

PR #543, opened 2024-11-04
Remove the skip from e2e test for getByPersistentId, so the test case could be included

### LINKED ISSUE #535: Implement the getDatasetCitation once it includes deaccessioned datasets
There is [an existing TODO](https://github.com/IQSS/dataverse-frontend/blob/8b59a48e985912986f8fd64ca8b1f1eef97d58c3/tests/e2e-integration/integration/datasets/DatasetJSDataverseRepository.spec.ts#L326). The test case is skipped because it's waiting the JS dataverse to add a check for deaccessioned dataset. Since the feature has been implemented, we could resolve this todo.

SPA users

ChengShi-1:
The test case of getByPersistentId should also work for deaccessioned datasets. The it.skip() should be removed because the JS dataverse has included deaccessioned api. 

- Closes #535 

It would be fine if e2e test DatasetJSDataverseRepository running well with no error message.

### DIFF
--- a/tests/e2e-integration/integration/datasets/DatasetJSDataverseRepository.spec.ts
+++ b/tests/e2e-integration/integration/datasets/DatasetJSDataverseRepository.spec.ts
@@ -336,6 +336,7 @@ describe('Dataset JSDataverse Repository', () => {
       const datasetExpected = datasetData(dataset.persistentId, dataset.version.id)
 
       expect(dataset.version.title).to.deep.equal(datasetExpected.title)
+      expect(dataset.version.publishingStatus).to.equal(DatasetPublishingStatus.DEACCESSIONED)
     })
   })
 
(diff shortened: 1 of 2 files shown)


tests/e2e-integration/integration/datasets/DatasetJSDataverseRepository.spec.ts
```diff
@@ -339,6 +338,7 @@ describe('Dataset JSDataverse Repository', () => {
       expect(dataset.version.title).to.deep.equal(datasetExpected.title)
     })
```
g-saracca, line 339: We are not testing that the dataset is actually deaccessioned.
I think there is a method to deaccesion a dataset, see DatasetHelper.deaccession method.

g-saracca (approved):

ofahimIQSS:
Test Ran successfully - merging pr

    ✔  All specs passed!                        01:22       12       11        -        1        -

PR #550, opened 2024-11-15
Upgrade to v4 download and upload artifacts

### LINKED ISSUE #537: Stop using v3 of download-artifact and upload-artifact
We need to update `download-artifact` and `upload-artifact` to latest 4 version.

`upload-artifact` is used in:

- deploy.yml
- deploy-beta-testing.yml
- test.yml

`download-artifact` is used in:

- deploy.yml
- deploy-beta-testing.yml

From https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/

"Starting November 30, 2024, GitHub Actions customers will no longer be able to use v3 of [actions/upload-artifact](https://github.com/actions/upload-artifact) or [actions/download-artifact](https://github.com/actions/download-artifact). Customers should update workflows to begin using [v4 of the artifact actions](https://github.blog/2024-02-12-get-started-with-v4-of-github-actions-artifacts/) as soon as possible. While v4 of the artifact actions improves upload and download speeds by up to 98% and includes several new features, there are [key differences](https://github.com/actions/upload-artifact?tab=readme-ov-file#breaking-changes) from previous versions that may require updates to your workflows. Please see [the documentation](https://github.com/actions/upload-artifact/blob/main/docs/MIGRATION.md) in the project repositories for guidance on how to migrate your workflows."

g-saracca:
Updates download and upload artifacts to version 4.
upload-artifacts have been updated in the following workflows:

- deploy.yml
- deploy-beta-testing.yml
- test.yml

download-artifacts have been updated in the following workflows:

- deploy.yml
- deploy-beta-testing.yml

- Closes #537 

Visual code inspection and re-run the **_test_** action. 
`deploy.yml` workflow has not been used.
`deploy-beta-testing.yml` workflow has been already tested deploying this current branch and it can be checked again after merging this branch into develop.

**_Note:_** the code coverage (coverall report) says it has decreased but it is the same rare problem that keeps occurring sometimes, we should not worry about it on this issue.

No

No

## Additional documentation:
No

### DIFF
--- a/.github/workflows/deploy-beta-testing.yml
+++ b/.github/workflows/deploy-beta-testing.yml
@@ -3,7 +3,7 @@ name: 'Deploy to Beta Testing'
 on:
   push:
     branches:
-      - develop
+      - 537-upgrade-to-v4-download-upload-artifacts
 
 jobs:
   build:
@@ -41,10 +41,11 @@ jobs:
       - name: Build with base path
         run: npm run build -- --base=/spa
 
-      - uses: actions/upload-artifact@v3
+      - uses: actions/upload-artifact@v4
         with:
           name: built-site
           path: ./dist
+          include-hidden-files: true
 
   deploy-to-payara:
     needs: build
@@ -58,7 +59,7 @@ jobs:
           distribution: 'zulu'
           java-version: '11'
 
-      - uses: actions/download-artifact@v3
+      - uses: actions/download-artifact@v4
         with:
           name: built-site
           path: ./dist
--- a/.github/workflows/deploy-beta-testing.yml
+++ b/.github/workflows/deploy-beta-testing.yml
@@ -3,7 +3,7 @@ name: 'Deploy to Beta Testing'
 on:
   push:
     branches:
-      - 537-upgrade-to-v4-download-upload-artifacts
+      - develop
 
 jobs:
   build:
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -61,10 +61,11 @@ jobs:
         if: ${{ github.event.inputs.infra_type == 'Payara' && github.event.inputs.basepath != '' }}
         run: npm run build -- --base=/${{ github.event.inputs.basepath }}
 
-      - uses: actions/upload-artifact@v3
+      - uses: actions/upload-artifact@v4
         with:
           name: built-site
           path: ./dist
+          include-hidden-files: true
 
   deploy-to-s3:
     needs: build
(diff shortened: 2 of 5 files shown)


stevenwinship (commented):
In dataverse project we had to use v4.1.7. Is there a reason not to use this instead of v4?

        # using v4.1.7 due to a bug in v4
        uses: actions/download-artifact@v4.1.7

g-saracca:
> In dataverse project we had to use v4.1.7. Is there a reason not to use this instead of v4?
Hi @stevenwinship, thanks! I thought using only v4 would always point to the latest version 4, but it might be safer to use an exact version, could you also tell me what exact version you are using for the upload-artifact?

pdurbin:
> I thought using only v4 would always point to the latest version 4

It's very confusing. https://github.com/orgs/community/discussions/25248#discussioncomment-3247083 by @ethomson (hi, Edward! 👋 ) says, "we only match a literal release tag".

However the GitHub Actions Toolkit docs at https://github.com/actions/toolkit/blob/master/docs/action-versioning.md#versioning say, "Binding to a major version is the latest of that major version ( e.g. v1 == "1.*" )" Then it goes on to say this:

"Make the new release available to those binding to the major version tag: Move the major version tag (v1, v2, etc.) to point to the ref of the current release."

So it might depend on if people are using GitHub Actions Toolkit or not? I'm not sure.

pdurbin:
@ethomson thanks for getting back to us so fast! Does this mean that for any action we can't know if the author is making updates to the "major version tag" or not? We have to poke around?

For example, when I look at the "download-artifact" action, I'm (now) fairly confident that "v4" means "latest 4" by looking at the following:

- The latest release is https://github.com/actions/download-artifact/releases/tag/v4.1.8 and came out on July 5th.
- If I manually navigate to https://github.com/actions/download-artifact/releases/tag/v4 ("v4" doesn't appear in the listing), the sha matches v4.1.8 (fa0a91b) as does the date (July 5th).

stevenwinship (dismissed):

ethomson:
> @ethomson thanks for getting back to us so fast! Does this mean that for any action we can't know if the author is making updates to the "major version tag" or not? We have to poke around?

I think that you do. And there's a few cases here: the author does this _always, and reliably_, the author never does this, or the author has a manual process to do this when they publish and maybe sometimes forgets.

I think that GitHub is pretty good about always updating the major version tag for their actions. (Hopefully automated, or at least tested; I honestly don't remember.) But me, personally, I'm in the latter camp and sometimes forget. 😢

ofahimIQSS:
It looks like test / component (pull_request) is failing - https://github.com/IQSS/dataverse-frontend/actions/runs/11862443295/job/33392252664?pr=550

ofahimIQSS (approved):
changes look good

PR #571, opened 2024-12-16
Fix Edit collection page type errors

### LINKED ISSUE #570: Typescript update
We need to update typescript version to understand Types Union props.
Also adding typecheck script to the lint script steps could be handy.

g-saracca:
This PR adds a typecheck script to the lint script.
It also updates Typescript to the latest version so it can Types Union Props that was causing the error.

- Closes #570 

1. Execute `npm i`.
2. Navigate with `cd packages/design-system && npm i && npm run build`.
3. Return with `cd ../../`.

**Step 2: Test the fix**
Change something in the code that will throw a type error and then run `npm run lint` to validate that shows the error.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "0.1.0",
       "dependencies": {
         "@faker-js/faker": "7.6.0",
-        "@iqss/dataverse-client-javascript": "2.0.0-alpha.6",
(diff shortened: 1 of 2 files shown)


ekraffmiller (approved):
There is a warning about the typescript version, but that can be a separate issue. Approved 👍

ofahimIQSS:
tested - looks good

PR #575, opened 2024-12-31
 Featured Items - Collection Page

### LINKED ISSUE #568: Featured Items - Collection Page

Add featured items section in the Collection Page.
Allows users that can edit a collection to create "featured items" that will initially support (content and an optional image).
These "featured items" will be shown between the Collection page header and the items list.
This issue its based on the initial analysis in https://github.com/IQSS/dataverse-frontend/issues/514.

Collection Owners

Q4 Roadmap

https://github.com/IQSS/dataverse/issues/10943
https://github.com/IQSS/dataverse-client-javascript/issues/202

g-saracca:
Add featured items section in the Collection Page.
Allows users that can edit a collection to create "featured items" that will initially support content and an optional image.
These "featured items" will be shown between the Collection page header and the items list.

Storybook links:
- https://646f68aa9beb01b35c599acd-nvjhuorxtc.chromatic.com/?path=/story/pages-collection--with-featured-items
- https://646f68aa9beb01b35c599acd-nvjhuorxtc.chromatic.com/?path=/story/pages-edit-collection-featured-items--default

- Closes #568

1. Execute `npm i`.
2. Navigate with `cd packages/design-system && npm i && npm run build`.
3. Return with `cd ../../`.

**Step 2: Test the feature**
1. Login and create a new collection or even in the root collection click the Edit Collection > Featured Items button.
2. Create featured items, with different contents, with and without image and submit the form, change the order of the featured items by dragging them with the drag handle.
3. Check that the featured items shown in the collection page are correct.
4. Go to edit them again, replace images from a featured item, also you can test removing an image from a featured item and also remove a featured item by clicking the minus button. *Note: you can remove the first item, to be removed needs to be at least in the 2nd position.
5. Check that the featured items shown in the collection page are correct.
6. Go to edit them again but now try the delete all featured items action, when clicking it, a confirmation dialog will appear.
7. Check in the collection page that no featured items are shown.
8. About field validations: the content field cannot be empty and must be 15,000 characters or less, but note that if you enter "Hello world" in the field, this rich text editor will transform that into `<p class="rte-paragraph">Hello world</p>` so those html tags characters also count for the validation.
9. You could also validate that if a collection is unpublished and has featured items, only users that can view unpublished collections can see the featured items.

Yes.
Screen Shot 2025-01-02 at 08 49 32
Screen Shot 2025-01-02 at 08 55 29
Screen Shot 2025-01-02 at 08 55 40


g-saracca:
Waiting for:

- https://github.com/IQSS/dataverse-client-javascript/issues/202
- https://github.com/IQSS/dataverse/issues/10943

src/App.tsx
```diff
@@ -15,9 +17,12 @@ if (BASE_URL === '') {
...
 function App() {
   return (
-    <SessionProvider repository={userRepository}>
-      <Router />
-    </SessionProvider>
+    <>
+      <SessionProvider repository={userRepository}>
+        <Router />
+      </SessionProvider>
+      <ToastContainer position="top-right" autoClose={5000} pauseOnHover />
```
ekraffmiller, line 24: I think the progress bar is a little confusing because it implies that the update action is still in process, what do you think?  Maybe we can have it display only for long running processes.
```suggestion
      <ToastContainer hideProgressBar={true}  position="top-right" autoClose={5000} pauseOnHover />
```

src/sections/collection/Collection.tsx
```diff
@@ -81,6 +89,12 @@ export function Collection({
                 {t('publishedAlert')}
               </Alert>
             )}
+            {showCollectionFeaturedItems && (
+              <FeaturedItems
+                collectionRepository={collectionRepository}
+                collectionId={collection.id}
+              />
+            )}
```
ekraffmiller, line 97: I tested the Collection page, and the user gets Page Not Found if they don't have permission to view the collection.  So is it ok to remove this? I think it would simplify the logic
```suggestion
              <FeaturedItems
                collectionRepository={collectionRepository}
                collectionId={collection.id}
              />
```

src/App.tsx
g-saracca: That's ok, when the toast appears the action was already done, that's why the text says 'sucessfully updated' or 'successfully deleted', the progress bar is only to show when the toast will go away.

src/sections/collection/Collection.tsx
g-saracca: Great, change applied already.

src/App.tsx
g-saracca: We can remove it if you think is better yes, but for long running processes will not add any value because as I said, when the toast appears the action was already completed.

ekraffmiller: ok, I see what you mean that it's not needed for long running processes. Maybe it's best to leave it for now and see what others think, it's easy to change later 👍

g-saracca: yes, is really easy to change and also those are global settings but you can override them from every individual toast you trigger

ekraffmiller (approved):
Looks good! I just found a small issue in the Rich Text Editor, that I opened another issue for.

ofahimIQSS:
trying to follow these steps but hit a snag on step 6 on building the environment

image

please advise :)

g-saracca:
@ekraffmiller could you please reapprove?

ekraffmiller (dismissed):
Approved version update & typo fixes

ofahimIQSS:
Are we setting any limits on how many Content items we can have...On the Edit Featured Items page I added 101 content items, when I did, I had the observation below:

https://github.com/user-attachments/assets/93094c82-6f61-48d5-836b-17734819b7af

To reproduce, just create a collection, edit> Featured Items, and add 100 Content items. The button class under the carousel seems to not function as expected. (not centered, tracker ball on button disappears after scrolling to beginning or end)

g-saracca:
> Are we setting any limits on how many Content items we can have...

Good question, I don't know about that, @GPortas what do you think? max 10 featured items for now?

ofahimIQSS:
another observation when there are too many featured items:
Page doesn't auto scroll down on local if I scroll up, I can keep hitting the "+" sign and in storybook, it looks like the page scrolls even if I scroll up a bit and add items. 

https://github.com/user-attachments/assets/c5627b9f-2b56-475d-be1d-04e65ad9ffea

g-saracca:
Thanks @ofahimIQSS, will take a look at this

ofahimIQSS:
merging is blocked - can someone please review

g-saracca:
> can we limit the message to show one at a time until one message "You can add up to 10 featured items." goes away?

I think we could, we could limit how many alerts to show at the same time but I think it's not our problem if the user keeps clicking the add button and also it doesn't break anything.
I would keep it simple for now, if we really find a UX issue with this I would address it but I think it's not an issue at this point. 
What do you think @ofahimIQSS ?

ofahimIQSS:
> > can we limit the message to show one at a time until one message "You can add up to 10 featured items." goes away?
> 
> I think we could, we could limit how many alerts to show at the same time but I think it's not our problem if the user keeps clicking the add button and also it doesn't break anything. I would keep it simple for now, if we really find a UX issue with this I would address it but I think it's not an issue at this point. What do you think @ofahimIQSS ?

Agreed! lets keep it as is - please approve the review so I can merge this PR. Thanks!

ekraffmiller (approved):
looks good, approved

ofahimIQSS:
Thank you guys! This feature looks great - Merging PR

PR #581, opened 2025-01-12
Fix UploadDatasetFiles Component test

### LINKED ISSUE #481: Race Condition in UploadDatasetFiles unit test
There seems to be a race condition in the component test, UploadDatasetFiles.spec.tsx. 
The error happens in 'restrict uploaded file' test:
https://github.com/IQSS/dataverse-frontend/actions/runs/10743959041/job/29799827812.

Screenshot 2024-09-06 at 4 23 28 PM
This error only happens intermittently and the test was successful on the second attempt.

ekraffmiller:
Updates UploadDatasetFiles.spec.tsx to avoid race condition errors.

- Closes #481 

I wasn't able to consistently produce this error, so it's hard to tell if this will definitively fix the bug. But there were multiple places where the test referred to a checkbox by the order on the page, not by an id, which could produce problems since the RestrictedModal and the UploadedFiles component both contain multiple checkboxes.
I added a test that the RestrictedModel is closed, and I added testids to multiple elements to make sure the correct checkboxes are being tested.

Review the code, run locally and check the Github actions.

No

No

### DIFF
--- a/src/sections/upload-dataset-files/uploaded-files-list/UploadedFiles.tsx
+++ b/src/sections/upload-dataset-files/uploaded-files-list/UploadedFiles.tsx
@@ -153,6 +153,7 @@ export function UploadedFiles({
                       <div>
                         <Form.Group.Checkbox
                           label={t('restricted')}
+                          data-testid={'restricted_checkbox_' + file.fileName}
                           id={'restricted-' + file.key}
                           checked={file.restricted}
                           onChange={(event: FormEvent<HTMLInputElement>) =>
--- a/src/sections/upload-dataset-files/uploaded-files-list/files-header/FilesHeader.tsx
+++ b/src/sections/upload-dataset-files/uploaded-files-list/files-header/FilesHeader.tsx
@@ -75,6 +75,7 @@ export function FilesHeader({
           label={''}
           aria-label={t('filesHeader.selectAll')}
           id={'select-all'}
+          data-testid="select-all-files-checkbox"
           checked={selectAllChecked}
           onChange={all}
         />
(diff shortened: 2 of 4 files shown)


src/sections/upload-dataset-files/uploaded-files-list/UploadedFiles.tsx
```diff
@@ -153,6 +153,7 @@ export function UploadedFiles({
                       <div>
                         <Form.Group.Checkbox
                           label={t('restricted')}
+                          data-testid={'restricted_checkbox_' + file.fileName}
```
g-saracca, line 156: Is there the possibility that could be multiple files with same fileName?
If so but that is not going to be the case for the tests thats ok, if not maybe we should use something else?

g-saracca (commented):
Looks good, just one question.

src/sections/upload-dataset-files/uploaded-files-list/UploadedFiles.tsx
ekraffmiller: If that happens, the form should rename the second uploaded file, for example in JSF:
Screenshot 2025-01-14 at 1 28 08 PM

But that I tried doing that in the SPA and the second upload of the file fails silently, rather than renaming. I will create an issue for that.

g-saracca: ok so for testing purposes I think its ok to use the fileName right?

ekraffmiller: Yes I think so 👍

g-saracca (approved):

ofahimIQSS:
Tests are passing, merging PR

PR #583, opened 2025-01-13
Update README.md

jggautier:
To the list of changes that the SPA introduces, this PR adds that Dataverses are referred to as collections, instead.

- I didn't open an issue for this.

This PR doesn't introduce a user interface change

I don't think a release note is needed for this change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -210,7 +210,7 @@ The environment is accessible through the following URLs:
 > The original JSF Dataverses/Datasets list on the home page uses normal paging buttons at the bottom of the list.
 > We have implemented infinite scrolling in this list, replacing the normal paging buttons, but the goal would be to be
 > able to toggle between normal paging and infinite scrolling via a toggle setting or button.
->
+> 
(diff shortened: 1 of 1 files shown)


g-saracca (approved):

pdurbin (commented):
I'll let the frontend team approve this but it looks good to me!

One thought is that we could also reword the `#### Dataverses/Datasets list` item above to talk about collections instead of dataverses. This could be in a separate PR if we wish.

Also, I'm not sure how this items are ordered but this new one (`#### Dataverses are Called Collections`) probably belongs at or at least near the top.

ChengShi-1 (approved):

g-saracca:
@pdurbin I've implemented your docs suggestions, take a look please and approve it if you think its ok. Thanks!

pdurbin (approved):
Looks good! Approved!

ofahimIQSS:
proof-read, merging pr

Issue #586, opened 2025-01-22
Dataset Page: Terms Tab (Read Only Version)


ekraffmiller:
### Overview of the Feature Request
For this feature, we will implement an initial version of the Dataset Page Terms Tab, without the Guestbook, as described in [SPA Features for Q1 2025](https://docs.google.com/document/d/17U4evH3tZ1FZ_plwBDG2S1QpD4eTsWui6fm_XvXKLJ4/edit?usp=sharing).

### JSF Screenshots of the terms Tab: 

**Screenshot 1 - Default License agreement, no restricted file, no TOU**

<img width="975" alt="Image" src="https://github.com/user-attachments/assets/d0b1189c-b70e-437c-9c33-4811d1563ba4" />

**Screenshot 2 - with one or more restricted files, and additional terms defined (in this case only "Data Access Place")**

<img width="953" alt="Image" src="https://github.com/user-attachments/assets/1654fdfd-1637-49ff-85eb-e783209ff438" />

**Screenshot 3 - with Custom Terms of use** 

<img width="957" alt="Image" src="https://github.com/user-attachments/assets/9f669000-d256-49e3-8cf2-13833690e4f1" />

PR #589, opened 2025-01-29
586 dataset terms tab

ekraffmiller:
Adds the Dataset Page Terms Tab in read only mode

#586 
- Closes #


g-saracca (changes requested):
Looks good, big work here 👏🏼 , some minor comments and suggestions.
I left some suggestions on the active tab thing, I think to improve the UX whenever the terms tab is active it should be scrolled into view, let me know if you need help with that.
Also the default storybook of PublishDatasetModal has customTerms, maybe should be a separate story?

src/sections/dataset/Dataset.tsx
```diff
@@ -78,7 +82,19 @@ export function Dataset({
   if (isDatasetLoading && !dataset) {
     return <DatasetSkeleton />
   }
+  const handleCustomTermsClick = () => {
+    setActiveTab('terms')
+    const newParams = new URLSearchParams(searchParams)
+    newParams.set('tab', 'terms')
+    // Update URL without reloading
+    navigate(`?${newParams.toString()}`, { replace: true })
```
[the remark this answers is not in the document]
ekraffmiller: You're right, it's not needed for the active tab, but it's replicating behavior in the JSF. The idea is that it makes it easier for users to get the url with the tab=terms, and use it as a link to the Custom Terms for whatever purpose they need (as an alternate to the License link.).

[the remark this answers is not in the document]
g-saracca: Ok, so we need to find a way for it to not refetch things again when updating a query param in the url, we can review this together if you want, I think I have an idea and know where the problem is.

packages/design-system/src/lib/components/tabs/Tabs.tsx
```diff
@@ -1,15 +1,35 @@
...
+  }, [defaultActiveKey])
+
+  const handleSelect = (k: string | null) => {
+    if (k) {
+      setKey(k)
+      if (onSelect) {
+        onSelect(k)
+      }
+    }
+  }
```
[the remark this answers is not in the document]
ekraffmiller: thanks, that worked!

src/sections/dataset/dataset-terms/DatasetTerms.tsx
```diff
@@ -0,0 +1,87 @@
...
+  } = useGetFilesCountInfo({
+    filesRepository,
+    datasetPersistentId,
+    datasetVersion
+  })
+  const restrictedFilesCount = filesCountInfo
+    ? numberOfRestrictedFiles(filesCountInfo.perAccess)
+    : 0
+  const termsOfAccessIsEmpty = Object.values(termsOfUse.termsOfAccess).every(
+    (value) => value === undefined || typeof value === 'boolean'
```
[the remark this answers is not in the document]
g-saracca: Sounds good, thanks. Yep lets keep the same value the api is returning. Resolving conversation.

src/sections/Route.enum.ts
```diff
@@ -19,6 +19,16 @@ export const RouteWithParams = {
...
   CREATE_COLLECTION: (parentCollectionId: string) => `/collections/${parentCollectionId}/create`,
+  DATASETS: (persistentId: string, version?: string, tab?: string): string => {
+    let url = `${Route.DATASETS}?${QueryParamKey.PERSISTENT_ID}=${persistentId}`
+    if (version) {
+      url += `&${QueryParamKey.VERSION}=${version}`
+    }
+    if (tab) {
+      url += `&${QueryParamKey.TAB}=${tab}`
+    }
+    return url
```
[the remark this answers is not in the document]
ekraffmiller: true, I will remove that

ekraffmiller:
thanks @g-saracca , I applied the requested changes 👍

packages/design-system/src/lib/components/tabs/Tabs.tsx
```diff
@@ -4,12 +4,15 @@ import { Tabs as TabsBS } from 'react-bootstrap'
...
 interface TabsProps {
   defaultActiveKey: string
+  onSelect?: (key: string | null) => void
 }
 
-function Tabs({ defaultActiveKey, children }: PropsWithChildren<TabsProps>) {
-  return <TabsBS defaultActiveKey={defaultActiveKey}>{children}</TabsBS>
+function Tabs({ defaultActiveKey, onSelect, children }: PropsWithChildren<TabsProps>) {
+  return (
+    <TabsBS onSelect={onSelect} activeKey={defaultActiveKey}>
```
g-saracca, line 12: Why not adding an extra prop activeKey and also defaultActiveKey, just as the original component accepts.
Something like:
`<TabsBS onSelect={onSelect} activeKey={activeKey} defaultActiveKey={defaultActiveKey}>`

src/sections/dataset/dataset-summary/DatasetSummary.tsx
```diff
@@ -1,26 +1,28 @@
 import { SummaryFields } from './SummaryFields'
 import { DatasetLicense, DatasetMetadataBlock } from '../../../dataset/domain/models/Dataset'
-import { License } from './License'
+import { SummaryLicense } from './SummaryLicense'
 import { MetadataBlockInfoRepository } from '../../../metadata-block-info/domain/repositories/MetadataBlockInfoRepository'
 
 interface DatasetSummaryProps {
   summaryFields: DatasetMetadataBlock[]
-  license: DatasetLicense
+  license: DatasetLicense | undefined
```
g-saracca, line 8: maybe `license?: DatasetLicense` instead?

src/sections/dataset/dataset-summary/SummaryLicense.tsx
```diff
@@ -1,12 +1,13 @@
-import { Row, Col } from '@iqss/dataverse-design-system'
+import { Row, Col, Button } from '@iqss/dataverse-design-system'
 import { DatasetLicense as LicenseModel } from '../../../dataset/domain/models/Dataset'
 import { useTranslation } from 'react-i18next'
 
 interface LicenseProps {
-  license: LicenseModel
+  license: LicenseModel | undefined
```
g-saracca, line 6: same here, `license?: LicenseModel` instead?

src/sections/dataset/dataset-terms/DatasetTermsRow.tsx
```diff
@@ -0,0 +1,26 @@
+import { Col, Row } from '@iqss/dataverse-design-system'
+import { QuestionMarkTooltip } from '@iqss/dataverse-design-system'
+import { MarkdownComponent } from '@/sections/dataset/markdown/MarkdownComponent'
+import styles from '@/sections/dataset/dataset-terms/DatasetTerms.module.scss'
+
+interface DatasetTermsRowProps {
+  title: string
+  tooltipMessage: string
+  value: string | undefined
```
g-saracca, line 9: Optional prop also here could be defined as `value?: string`

src/sections/dataset/dataset-terms/EditDatasetTermsButton.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+export function EditDatasetTermsButton() {
+  const { t } = useTranslation('dataset')
+  const { user } = useSession()
+  const { dataset } = useDataset()
+  const { showModal } = useNotImplementedModal()
+
+  if (!user || !dataset?.permissions.canUpdateDataset) {
+    return null
+  }
```
g-saracca, line 17: If you return null here you will be still rendering an empty div in `DatasetTerms.tsx`.
Instead you could call the useDataset hook in `DatasetTerms.tsx` and conditionally render the whole div with the `EditDatasetTermsButton` inside, and also as you have the dataset info in `DatasetTerms.tsx` you could pass a prop called disabled based on `dataset.checkIsLockedFromEdits(user.persistentId)`, so you will need to move both hooks useSession and useDataset to the parent component.
What do you think? I prefer not render empty divs, because even if you dont render the inner component if the div has a margin or a padding it will still change how the UI looks.

g-saracca (changes requested):
Just some more minor things

src/sections/dataset/dataset-terms/EditDatasetTermsButton.tsx
ekraffmiller: @g-saracca I see what you mean about the empty div.  I was trying to follow a convention we have in other buttons, where the logic for displaying the button is within the component itself, rather than the parent. If you search for 'return <></>' in the code base, you can see them.  If I add the div container inside of the EditDatasetTermsButton, would that solve the problem?

g-saracca: yep that would indeed solve the problem also 👍🏼

ekraffmiller: great, I will do that

packages/design-system/src/lib/components/tabs/Tabs.tsx
ekraffmiller: I added the activeKey as suggested...I thought it was confusing to have all the props be optional, so I added some warnings inside the component, highlight how the props should be set, let me know what you think.

g-saracca (approved):
Looks good!

src/sections/shared/form/EditCreateCollectionForm/collection-form/useSubmitCollection.ts
```diff
@@ -82,9 +84,7 @@ export function useSubmitCollection(
       contacts: contactsDTO,
       metadataBlockNames: metadataBlockNamesDTO,
       inputLevels: inputLevelsDTO,
-      facetIds: facetIdsDTO,
-      inheritMetadataBlocksFromParent: useMetadataFieldsFromParentChecked,
-      inheritFacetsFromParent: useFacetsFromParentChecked
+      facetIds: facetIdsDTO
     }
```
g-saracca, line 88: Could you move this two at the bottom of the object as it is in develop branch?
```
inheritFacetsFromParent: useFacetsFromParentChecked,
inheritMetadataBlocksFromParent: useMetadataFieldsFromParentChecked,
```

ekraffmiller: @g-saracca  ok, moved the fields

g-saracca: thanks 🫶

g-saracca (approved):

ofahimIQSS:
Merging PR - tests are now passing

PR #609, opened 2025-02-21
Contact Collection and Dataset owner feature

### LINKED ISSUE #569: implement Collection Page: Contact button and form
The goal is to replicate the Contact feature on the Collection Page. 
- contact button with tooltip
- contact form
- CAPTCHA mechanism 
- notification

image
image
Screenshot 2024-12-12 at 1 33 15 PM

#567

ChengShi-1:

The goal is to replicate the Contact feature on the Collection Page and Dataset Page.
- contact button with tooltip
- contact form
- CAPTCHA mechanism
- notification

- Closes #569 

Not sure contact factory is written in a correct way or not to create contact repository

1. Perform the following tests in both Login and Logout states: If not logged in, there is an input box for emails
2. Verify that the Contact button:
- Exists on the page.
- Is clickable.
3. Validate form submission:
- Display errors and prevent submission if:
  - Required fields are empty.
  - The email format is incorrect.
  - The provided answer is incorrect.
4. Allow form submission if all inputs are correct:
Show a success alert on the page after submission.
image
image

No


packages/design-system/src/lib/components/alert/Alert.tsx
```diff
@@ -29,10 +30,15 @@ function Alert({ variant, dismissible = true, customHeading, children }: AlertPr
...
+  const handleClose = () => {
+    onClose && onClose()
+    setShow(false)
+  }
+
   return (
     <>
       {show && (
-        <AlertBS variant={variant} onClose={() => setShow(false)} dismissible={dismissible}>
+        <AlertBS variant={variant} onClose={handleClose} dismissible={dismissible}>
```
g-saracca, line 41: Could you avoid all these changes here and show a toast alert instead of showing an alert with this component? 
Just calling `toast.success('message here')` after the email is sent successfully is enough, less boilerplate.

public/locales/en/contact.json
```diff
@@ -0,0 +1,26 @@
+{
```
g-saracca, line 1: Could you place all this translations in the shared.json file, maybe under a property with the same name as the component name?

src/contact/domain/repositories/ContactRepository.ts
```diff
@@ -0,0 +1,6 @@
+import { ContactDTO } from '../useCases/ContactDTO'
+import { Contact } from '../models/Contact'
+
+export interface ContactRepository {
+  submitContactInfo: (contactDTO: ContactDTO) => Promise<Contact[]>
```
g-saracca, line 5: I found this `submitContactInfo` naming a little confusing, I saw in js-dataverse is the same.
This methods suggest you are submitting some contact information but what you are really doing is sending feedback to the owners right?

src/contact/domain/useCases/submitContact.ts
```diff
@@ -0,0 +1,13 @@
+import { ContactRepository } from '../repositories/ContactRepository'
+import { WriteError } from '@iqss/dataverse-client-javascript'
+import { ContactDTO } from './ContactDTO'
+import { Contact } from '../models/Contact'
+
+export async function submitContact(
```
g-saracca, line 6: Same here about the naming and also the ContactDTO naming

src/sections/contact/ContactButton.tsx
```diff
@@ -0,0 +1,73 @@
+import { useState } from 'react'
```
g-saracca, line 1: Directly under sections folder should be only pages (collection, datasets, files, etc) and contexts, that is how it has been done ( not sure why contexts but has been done like that).
So this ContactButton could live in sections/shared folder and the ContactModal if it is always going to be used with this button should live inside this component also.
Check in this PR the `delete-collection-button` folder.

src/sections/contact/ContactButton.tsx
```diff
@@ -0,0 +1,73 @@
...
+import { useTranslation } from 'react-i18next'
+import { Button, Tooltip } from '@iqss/dataverse-design-system'
+import { Envelope } from 'react-bootstrap-icons'
+import { ContactModal } from '@/sections/shared/contact-modal/contact-modal'
+import { ContactRepository } from '@/contact/domain/repositories/ContactRepository'
+
+interface ContactButtonProps {
+  onSuccess: () => void
+  toContactName: string
+  isCollection: boolean
```
g-saracca, line 11: Instead of a boolean with isCollection you can use union types, so you are not tied to only two possible options.
ContactObjectType: 'collection' | 'dataset' | 'file' for example.
Then I would keep the buttons styles the same so it is consistent across the app.

src/sections/contact/contactFactory.tsx
```diff
@@ -0,0 +1,8 @@
+import { ContactRepository } from '@/contact/domain/repositories/ContactRepository'
+import { ContactJSDataverseRepository } from '@/contact/infrastructure/ContactJSDataverseRepository'
+
+export class ContactRepositoryFactory {
+  static create(): ContactRepository {
+    return new ContactJSDataverseRepository()
+  }
+}
```
g-saracca, line 8: Factories are only used for pages components not for repositories.

src/sections/dataset/Dataset.tsx
```diff
@@ -111,13 +113,18 @@ export function Dataset({
               <Row>
                 <Col>
                   <DatasetAlerts />
+                  {contactSuccess && (
+                    <Alert variant="success" onClose={() => setContactSuccess(false)}>
+                      {tcontact('contact.contactSuccess')}
+                    </Alert>
+                  )}
```
g-saracca, line 120: You can remove this Alert and the useState to handle the message after using toast.success from 'react-toastify' package.

src/sections/dataset/Dataset.tsx
```diff
@@ -111,13 +113,18 @@ export function Dataset({
...
                 </Col>
               </Row>
             </div>
 
             <header className={styles.header}>
               <h1>{dataset.version.title}</h1>
-              <DatasetLabels labels={dataset.version.labels} />
+              <DatasetLabels labels={dataset.version.labels} />{' '}
```
g-saracca, line 127: Why the empty string?

src/sections/dataset/dataset-action-buttons/DatasetActionButtons.tsx
```diff
@@ -10,19 +10,24 @@ import { EditDatasetMenu } from './edit-dataset-menu/EditDatasetMenu'
...
 
 export function DatasetActionButtons({
   dataset,
   datasetRepository,
-  collectionRepository
+  collectionRepository,
+  onSuccess
 }: DatasetActionButtonsProps) {
   const { t } = useTranslation('dataset')
+  const contactRepository = ContactRepositoryFactory.create()
```
g-saracca, line 30: Please don't cerate repositories with factories and not inside components otherwise every time the component re renders you will create a new instance of that class.
All repositories are created in the pages factories ( DatasetFactory, CollectionFactory and so on) and pass as props down the tree.

src/sections/shared/contact-modal/contact-modal.tsx
```diff
@@ -0,0 +1,114 @@
+import { useEffect } from 'react'
```
g-saracca, line 1: Move this contact-modal inside contact-button component according to comment above.
Also contact button maybe could have a more explicit name like ContactOwnerButton or something like that.

g-saracca (changes requested):
I leave several suggestions to improve the code and organize it in a better way.
I think the way you name the contact repository methods is a bit confusing, `ContactDTO`, `submitContactInfo`, `useSubmitContact`. Maybe something more related to what you are doing which is sending feedback to the owners.
On the other hand I think you had a confusion between repositories and factories, factories is a convention that the project has to create or instantiate the components at page level.
And repositories must be created only once in these factories and passed through props to the component that wants to use it.

Also you can simplify success message with a toast instead of an alert.

Please take a look at all the comments first and then see what needs to be changed.
If you need any help let me know, after these changes I will analyze the UI and functionality in more depth.

ChengShi-1:
@g-saracca Hi German, thanks for your feedback. I made changes regarding to reviews. Please help to check them again, ty!

src/sections/shared/contact/contact-modal/contact-modal.tsx
```diff
@@ -0,0 +1,111 @@
...
+        {submissionStatus === SubmissionStatus.Errored && (
+          <Alert variant="danger" dismissible={false}>
+            {submitError}
+          </Alert>
+        )}
+        <FormProvider {...methods}>
+          <form onSubmit={methods.handleSubmit(onSubmit)}>
+            <ContactForm isLoggedIn={Boolean(user)} toContactName={toContactName} />
+          </form>
+          <Captcha />
```
g-saracca, line 91: `<Captcha/>` could live inside `<ContactForm/>`?

src/sections/shared/form/ContactForm/useSendFeedbacktoOwners.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+  const submitForm = async (formData: FeedbackDTO): Promise<Contact[] | string> => {
+    setSubmissionStatus(SubmissionStatus.IsSubmitting)
+    setSubmitError(null)
+    try {
+      const contacts: Contact[] = await contactRepository.sendFeedbacktoOwners(formData)
+      setSubmissionStatus(SubmissionStatus.SubmitComplete)
+      return contacts
+    } catch (error) {
+      setSubmitError(error instanceof Error ? error.message : 'Unknown error occurred')
```
g-saracca, line 35: This error handling is ok, but i found a better way check [DeleteFile PR useDeleteFile hook](https://github.com/IQSS/dataverse-frontend/pull/606/files#diff-bda39d91a86c9068efca92e4b9311a17df2f188e1d2b1297e9942396b21e32b7)

tests/component/sections/dataset/dataset-action-buttons/DatasetActionButtons.spec.tsx
```diff
@@ -56,6 +59,7 @@ describe('DatasetActionButtons', () => {
         dataset={dataset}
         datasetRepository={datasetRepository}
         collectionRepository={collectionRepository}
+        onSuccess={() => {}}
```
g-saracca, line 62: Is this prop still needed?

tests/component/sections/shared/pagination/contact-modal/ContactModal.spec.tsx
```diff
@@ -0,0 +1,84 @@
...
+        show
+        title={title}
+        handleClose={() => {}}
+        toContactName={toContactName}
+        id="123"
+        contactRepository={contactRepository}
+      />
+    )
+  })
+  it('should should fields and information correctly', () => {
```
g-saracca, line 29: double should

packages/design-system/src/lib/components/alert/Alert.tsx
```diff
@@ -8,7 +8,7 @@ interface AlertProps {
   variant: AlertVariant
   dismissible?: boolean
   customHeading?: string
-  children: ReactNode
+  children?: ReactNode
```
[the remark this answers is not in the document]
ChengShi-1: I am not sure if it's necessary for every alert, because some <Alert> in using don't have a `children` there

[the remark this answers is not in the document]
g-saracca: oh didn't know that, could you show me an example? because if we are using an <Alert component without a children we should have lint errors.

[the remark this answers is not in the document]
ChengShi-1: I met lint error when I add Alert without children, but there is no error on my side for those existed already. [like CollectionForm](https://github.com/IQSS/dataverse-frontend/blob/develop/src/sections/shared/form/EditCreateCollectionForm/collection-form/CollectionForm.tsx)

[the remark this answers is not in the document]
g-saracca: oh but that has a children which is`{submitError}`.
`children` prop in react is not a custom prop, is a built in prop that will have the value of what you pass in between open and closing tags of a component.
`<MyComponent><p>I will be read as children</p></MyComponent>`

See react docs https://18.react.dev/learn/passing-props-to-a-component#passing-jsx-as-children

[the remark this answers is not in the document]
ChengShi-1: ohhh Thanks for explaining! I misunderstand it, I thought it was a prop

src/contact/domain/models/Contact.ts
```diff
@@ -0,0 +1,5 @@
+export interface Contact {
```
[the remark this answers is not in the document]
ChengShi-1: It has a little bit difference, the reply message's body would be `Hello ${User Name},\n\nYou have just been sent the following message from ${FromEmail} via the Root hosted dataset ...` but I didn't use the reply message in Contact part in front end repo.

ChengShi-1:
> Some more comments, thanks for addressing previous requested changes.

Changes applied. Thanks for taking care of this PR

g-saracca (dismissed):
Looks great! Approving, nice work!

g-saracca:
I've approved but I saw that component test  coverage is not reaching the branch coverage min of 95%.
Please check what code branches you are not testing, thanks, almost there! 🤞🏼

g-saracca:
I think you are missing testing closing the dialog? See https://coveralls.io/builds/72453589/source?filename=src%2Fsections%2Fshared%2Fcontact%2Fcontact-modal%2Fcontact-modal.tsx#L74

g-saracca (approved):
Looks good! Approving!

ofahimIQSS:
tested updating User email successfully:

https://github.com/user-attachments/assets/2c5f3b8c-9bdc-4121-b6a3-eb53f42b09fb

g-saracca (approved):

ofahimIQSS:
Looks good from my end, merging PR

PR #627, opened 2025-03-19
use a semicolon as the separator when displaying multiple metadata field values

### LINKED ISSUE #377: Dataset Page: separator character missing when displaying multiple metadata field values
For displaying simple values that have multiples, for example Subject, there should be a separator character between values
JSF Example:
Screenshot 2024-04-17 at 3 52 16 PM
 
 Currently in the SPA, we don't have a separator character because the logic is using the "displayFormat" field of the TSV. The  Instead of doing that, we should do the formatting in the component.
Screenshot 2024-04-17 at 3 51 31 PM

pdurbin:

Previously, there was no separator, as shown in the issue:

Screenshot 2025-03-19 at 4 20 48 PM

- Closes #377

As suggested in the issue, I removed any reliance on displayFormat from the TSV and simply hard-coded a semicolon.

I tested "Subject" (and added a Cypress test for this) but it would be good to test other fields that can have multiple values.

Yes, not the semicolon between subjects:

Screenshot 2025-03-19 at 4 13 10 PM

No.

## Additional documentation:

None.

### DIFF
--- a/src/sections/dataset/dataset-metadata/dataset-metadata-fields/DatasetMetadataFieldValueFormatted.tsx
+++ b/src/sections/dataset/dataset-metadata/dataset-metadata-fields/DatasetMetadataFieldValueFormatted.tsx
@@ -58,7 +58,7 @@ export function metadataFieldValueToDisplayFormat(
   metadataFieldValue: DatasetMetadataFieldValueModel,
   metadataBlockInfo?: MetadataBlockInfoDisplayFormat
 ): string {
-  const separator = metadataBlockInfo?.fields[metadataFieldName]?.displayFormat ?? ''
+  const separator = ";"
 
   if (isArrayOfObjects(metadataFieldValue)) {
     return metadataFieldValue
--- a/tests/e2e-integration/e2e/sections/create-dataset/CreateDataset.spec.tsx
+++ b/tests/e2e-integration/e2e/sections/create-dataset/CreateDataset.spec.tsx
@@ -33,6 +33,7 @@ describe('Create Dataset', () => {
         cy.findByLabelText('Toggle options menu').click({ force: true })
 
         cy.findByLabelText('Agricultural Sciences').click()
+        cy.findByLabelText('Arts and Humanities').click()
       })
     cy.findByText(/Save Dataset/i).click()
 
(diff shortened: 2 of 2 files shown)


g-saracca (changes requested):
Works good!
I think now we just need to remove the unused `metadataFieldName` argument from the  `metadataFieldValueToDisplayFormat` function.

I can see that this function it is used in:
- In the same component where it is declared.
- In `DatasetMetadata.spec.tsx `
- In `SummaryFields.spec.tsx`

pdurbin:
@g-saracca good catch! In f32c8b5 I removed those references. Thanks.

On a related note, I'm using VS Code (but I'm open to other options) and "Find All References" (which is like "Find Usages" in IntelliJ) couldn't find the references in the test directories without a change like this:

```
% git diff
diff --git a/tsconfig.json b/tsconfig.json
index 7a25b67d..5d3f4fe0 100644
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -30,5 +30,5 @@
     "packages/design-system/src/",
     "public"
   ],
-  "exclude": ["tests/**/**.spec.ts", "tests/**/**.spec.tsx"]
+  "exclude": []
 }
```

Should I open an issue or a PR about this? (Here's a related post: https://stackoverflow.com/questions/52677651/find-all-references-across-files-for-a-method-function-in-vscode-is-not-possib/52711331#52711331 .) I assume we exclude them for a reason! 😅 

Finally, is there a way to run just a single unit test? Or test from a single file, at least? I'm happy to make a PR to explain how if you tell me. 😄 (I see we explain how for Cypress tests, which is great.)

g-saracca (approved):
Hi phil, thanks for the changes, approving!

g-saracca:
Hi phill, I work with VsCode also and I search words with the search bar on the sidebar, I'm used to that, it works just fine.

About `tsconfig` file  I think that excluding the tests files is a little bit weird and hide some ts problems for us. I will try to find out what was the reason for that and if there was not one I will put it back to `[]`.
Please open an issue to investigate that.

About the Chromatic UI tests, no, the person doing the code review should also check the new stories and approve them if ok.

ofahimIQSS:
UI Review: dataverse-frontend still on waiting for status to be reported.

g-saracca:
> UI Review: dataverse-frontend still on waiting for status to be reported.

@ofahimIQSS , Done.

ofahimIQSS:
reproduced issue in beta:
I filled out all the metadata fields with multiple values:
image

ofahimIQSS:
Alrighty, tested the PR in local, please see the screenshot for the SPA - I jotted down all my observations in red. Please let me know if I should open separate PRs for this or if you want to handle all that in this and modify the title of pr? 
SPA:
image

JSF:
image

g-saracca:
@pdurbin @ofahimIQSS 
I think hardcoding the separator as `;` maybe was not the correct decision, I didn't test it locally when code reviewing, but this is what I see now on my local environment previous to this PR and it's seems fine, except for the separators for multiple field values.

Screen Shot 2025-03-27 at 09 48 52

Issue #639, opened 2025-03-31
File Page: File Restrict Access Request & Terms of Access


ChengShi-1:
**Overview of the Feature Request**
In edit file restriction feature, we're missing File Restriction Access Request & Terms of Access

**What kind of user is the feature intended for?**
(Example users roles: API User, Curator, Depositor, Guest, Superuser, Sysadmin)

**What inspired the request?**

**What existing behavior do you want changed?**

**Any brand-new behavior do you want to add to Dataverse?**

**Any open or closed issues related to this feature request?**
https://github.com/IQSS/dataverse-client-javascript/pull/279

PR #641, opened 2025-04-02
File Page: Restrict Modal update restrict access and terms

ChengShi-1:
In edit file restriction feature, we're missing File Restriction Access Request & Terms of Access

- Closes #https://github.com/IQSS/dataverse-frontend/issues/639

npm i 

Please test it on both file page and dataset page > File tab for every single file. 
1. The checkbox of enable access should work.
2. The terms of access should be able to type and edit.
3. If a published dataset enables access, change to another account (not the dataset owner), Access Request button should be shown.

Also, check some rules below:
If restrict is true and enableAccessRequest is false then termsOfAccess is required.
The enableAccessRequest and termsOfAccess are applied to the Draft version of the Dataset and affect all of the restricted files in said Draft version.

image
Screenshot 2025-04-02 at 7 33 56 PM


src/dataset/infrastructure/repositories/DatasetJSDataverseRepository.ts
```diff
@@ -108,7 +108,7 @@ export class DatasetJSDataverseRepository implements DatasetRepository {
       DatasetNonNumericVersion.LATEST_PUBLISHED,
       DatasetNonNumericVersion.DRAFT
     ).then((datasetVersionDiff) => {
-      datasetDetails.datasetVersionDiff = datasetVersionDiff
+      datasetDetails.datasetVersionDiff = datasetVersionDiff as JSDatasetVersionDiff
```
g-saracca, line 111: You shouldn't cast this here with the `as JSDatasetVersionDiff`, JSDatasetVersionDiff now has a new property called versionState inside VersionSummary.
versionState is in js-dataverse equal to this:
```
export declare enum DatasetVersionState {
    DRAFT = "DRAFT",
    RELEASED = "RELEASED",
    ARCHIVED = "ARCHIVED",
    DEACCESSIONED = "DEACCESSIONED"
}
```
Please add this property to the spa DatasetVersionDiff model and that's it ;), no more casting, let me know if you have any doubt please.
You can place this mentioned enum in the Dataset model inside the spa.

src/files/domain/useCases/restrictFileDTO.ts
```diff
@@ -0,0 +1,5 @@
+export interface RestrictDTO {
```
g-saracca, line 1: Please rename as same as file name, to `RestrictFileDTO`

src/sections/file/file-action-buttons/edit-file-menu/restrict-file-button/useRestrictFile.tsx
```diff
@@ -27,11 +32,24 @@ export const useRestrictFile = ({
...
+    enableAccessRequest: boolean,
+    terms?: string
+  ) => {
     setIsRestrictingFile(true)
 
+    const restrictDTO: RestrictDTO = {
+      restrict: !isRestricted
+    }
+
+    if (isRestricted == false) {
```
g-saracca, line 46: please use === instead of ==

g-saracca (changes requested):
Some requested changes.

Also some UI points:
1. The link to the guides should use latest instead of fixed version:
"https://guides.dataverse.org/en/6.5/user/..." to "https://guides.dataverse.org/en/latest/user/..."
Also you should use a normal `<a/>` tag with a target _blank so it opens the link in a new tab and include the link text in the translations with <Trans component. See[ this example](https://github.com/IQSS/dataverse-frontend/blob/51f1d8489236e4fe51de2edeb47b4b7f305eaa20/src/sections/shared/file-uploader/file-upload-input/FileUploadInput.tsx#L250).
2. Add `<QuestionMarkTooltip />` to "Request Access" and "Terms of Access for Restricted Files" labels, see JSF version of the modal. There are lot of examples of how to use it within the SPA code.
3. Reset form fields values when closing the modal.

ofahimIQSS:
Looks good, merging:

image
image

PR #648, opened 2025-04-08
Send only input levels that changed

### LINKED ISSUE #628: Create/Edit collection - send only input levels that changed
When creating a collection or an input level, we need to only send input levels that changed from its original value so we don't save unneeded data in db.

g-saracca:
When creating a collection or an input level, we need to only send input levels that changed from its original value so we don't save unneeded data in db.

- Closes #628 

Check that create and edit collection functionality keeps working correctly.
This is just a change to send only the input levels that were modified from the original form value when editing or creating a collection.

No

No

## Additional documentation:
No


ChengShi-1 (approved):
looks good

PR #672, opened 2025-04-29
User Feedback CTA

### LINKED ISSUE #663: Research/Implement User Feedback CTA
Research and implement a call-to-action component to gather user feedback within the SPA.
The idea is to have a floating button at the bottom left of the screen for users to give feedback about different parts of the SPA.
As an initial step this will only add the UI side of it, without actually sending the feedback to an API, but we will also do some research on how to do this and what api extensions we will need for it.

g-saracca:
This just adds the UI for sending feedback from the SPA.
It's still not integrated with any js-dataverse use case nor dataverse API.

- Closes #663 

Step 1: Run the Development Environment

1. Execute npm i.
2. Navigate with cd packages/design-system && npm i && npm run build.
3. Return with cd ../../.
4. Ensure you have a .env file similar to .env.example, with the variable VITE_DATAVERSE_BACKEND_URL=http://localhost:8000/.
5. Navigate with cd dev-env.
6. Before running next step, be sure have a fresh unstable dataverse image.
7. Start the environment using ./run-env.sh unstable.
8. To verify the environment, visit http://localhost:8000/ and check your local Dataverse installation.

Step 2: Test the feature

This is just UI work, so it won't actually send any feedback yet.
What can be tested is:
- Validate that the CTA button is only shown when you are logged in.
- Validate that if you open the form in the Home page then in the form the "Home" option on the "Page" field is preselected, same should happen with any page within the SPA.
- Validate that when submitting the form without entering any feedback an error message is shown.
- Validate that the dialog closes with the "Cancel" button, the "X" icon button, pressing "Esc" key and clicking outside the dialog.
- The submission is mocked, so if you click the Send button you should see a success notification.

Yes.

Screenshot 2025-04-28 at 10 47 53
Screenshot 2025-04-29 at 17 16 19


g-saracca:
Adding waiting tag, we need more discussion on how to tackle this.

PR #687, opened 2025-05-08
Feat/646 links underlined style

### LINKED ISSUE #646: Links & Buttons type link underlined style
At stefano's request, we have to see if we can avoid links and buttons type link with the underlined style but keep being compliant with accessiblity.
The css to do this it's pretty straightforward like below, but when removing it there are violations in storybook accesibility check.
```

a,
.btn-link {
  text-decoration: none;
}
```

g-saracca:

- Closes #646

### DIFF
--- a/packages/design-system/.gitignore
+++ b/packages/design-system/.gitignore
@@ -25,3 +25,6 @@ dist-ssr
 
 # Tests
 coverage
+
+# Storybook
+storybook-static
--- a/packages/design-system/src/lib/assets/styles/bootstrap-customized.scss
+++ b/packages/design-system/src/lib/assets/styles/bootstrap-customized.scss
@@ -144,3 +144,8 @@ a,
 .btn-link {
   text-decoration: none;
 }
+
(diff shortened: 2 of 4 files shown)


ChengShi-1 (dismissed):
looks good

ofahimIQSS:
not sure whats going on but in my local, I get error pages when i try to create a collection and publish a dataset.
Might be because of my local setup, might need some guidance there. 

image

ofahimIQSS:
just had to delete browser data, then I got things cooking. 

What about these links that are displaying the underlines? 

image

g-saracca:
@ChengShi-1 could you please review again? Instead of changing underlined style from all `<a/>` elements now I'm targeting some specific elements that don't have any problem regarding a11y.
I've updated the PR description.
cc: @ofahimIQSS

ChengShi-1 (approved):

ofahimIQSS:
Looks good to me - merging

PR #688, opened 2025-05-08
Support for Shibboleth external users provided using Keycloak and integrated into the development environment

GPortas:

This PR enables the development environment to authenticate users via a Shibboleth Identity Provider (IdP) through Keycloak, and generate OIDC credentials for them to use within the single-page application (SPA). This addresses the following design that we defined a while ago for SAML-OIDC brokering:

421995656-b744306d-37f9-4643-b1da-e5c33176cb7f (1)

This PR introduces a fully dockerized Shibboleth Identity Provider (IdP), backed by a dockerized LDAP directory containing test users. Keycloak has been configured to operate as a SAML Service Provider (SP) for the Shib IdP, enabling federated authentication flows within the development environment.

This setup provides a reusable boilerplate configuration that will serve as the foundation for integrating HarvardKey authentication in the beta Keycloak deployment. While additional ad-hoc configuration will be required to interoperate with InCommon in production, this PR lays the groundwork by including the necessary base setup.

Since this PR introduces new test flows that might raise some questions, I’ve recorded an informal demo to walk through the functionality and highlight a few key points. The focus is mainly on how it works in practice. We can go over the technical decisions and justifications at our next tech hours meeting.

https://github.com/user-attachments/assets/4e700593-2356-472e-843d-4e2775ff7314

The dev-env reverse proxy has been updated to use HTTPS. With these changes, all services are now exposed under https://localhost/, using a development SSL certificate. Using HTTPS was necessary to avoid incompatibilities with the new authentication flows, and applying it to all services has significantly simplified their configuration.

Below is a more detailed diagram, and more related to the implemented changes:

Screenshot 2025-05-09 at 22 56 08

- Closes https://github.com/IQSS/dataverse/issues/11337

Initially, I created the PR in the backend repository, but later realized that adding Shib to the containerized environment is more relevant in the frontend repo, since that’s where we’ll be testing the end-to-end auth flows, including SPA + PKCE. However, if it becomes necessary to add Shib to the dataverse backend repo in the future, we can base it on what we have on the dataverse-frontend repo.

Follow the demo use cases.

N/A

N/A


qqmyers (commented):
Just looking at the design level, I think this makes general sense and is consistent with the way Keycloak brokers for OIDC (Google/ORCID logins) as well - they all appear as alternate methods. Similarly, the way linking works when there's a conflict makes sense from the Keycloak design and security side - as in the video, you can link if you can login to the existing account too.

I/we may want to think more about that from the Dataverse perspective though - we currently treat builtin and shib users differently, and base shib groups on how you log in, but don't really have the idea of one account where the user can login multiple ways. I think the key thing related to shib groups is that we want/need a way to group people by 'affiliation' where we know the affiliation is not user editable. 
Other questions for production use would be how to handle existing shib accounts - presumable with Keycloak in place, these now appear as OIDC users from the Keycloak provider - so how could we match up/migrate the existing accounts? (Are conflicts with existing shib users detected? or only those with builtin users?)
As @GPortas said - might be a good tech hour discussion to brainstorm and be able to ask deeper questions. My guess overall is that this PR is probably a good starting point, i.e. we'll have to add to or tweak it to get to a production capability, but I don't have a sense that there's an alternate approach (maybe someone else can see one?).

g-saracca (approved):
Looks good and is really isolated as it proposes a separate dev environment for shib stuff.

ofahimIQSS:
When I built this PR in local and hit login from the SPA - I'm not seeing the SAML option

image

g-saracca:
hi @ofahimIQSS, please check the steps from the `shib-dev-env` folder README - **_Setup Instructions_** section.
Run the `./run-env.sh unstable` command from inside the `shib-dev-env` folder also.

PR #700, opened 2025-05-16
File Page: File Version Summary Integration

### LINKED ISSUE #678: File Page: File Version Summary Integration
Implement the File Page > Version tab

- Table with four columns: dataset version, summary, contributors, published on
- Version number
- Summary( File(Added, Replaced, Deleted, Changes), File Metadata(Added, Replaced, Deleted, Changes), File Tags with count, Deaccessioned Reason, File not included)
- Contributors, only available once the file is edited
- Published on, format YYYY-MM-DD

Image

SPA User

https://github.com/IQSS/dataverse-client-javascript/issues/297

ChengShi-1:
In file page, we need a version tab showing the file changes in each dataset version.
File Page > Version tab

- Table with four columns: dataset version, summary, contributors, published on
- Version number: should be a button, link to available file 
- Summary( File(Added, Replaced, Deleted, Changes), File Metadata(Added, Replaced, Deleted, Changes), File Tags with count, Deaccessioned Reason, File not included)
- Contributors, 
- Published on, format YYYY-MM-DD

- Closes #678


src/files/infrastructure/FileJSDataverseRepository.ts
```diff
@@ -246,7 +251,8 @@ export class FileJSDataverseRepository implements FileRepository {
           FileJSDataverseRepository.getDownloadCountById(jsFile.id, jsFile.publicationDate),
           FileJSDataverseRepository.getPermissionsById(jsFile.id),
           FileJSDataverseRepository.getThumbnailById(jsFile.id),
-          FileJSDataverseRepository.getTabularDataById(jsFile.id, jsFile.tabularData)
+          FileJSDataverseRepository.getTabularDataById(jsFile.id, jsFile.tabularData),
+          getFileVersionSummaries.execute(jsFile.id)
```
g-saracca, line 255: This is what we probably want to avoid, fetching file version summaries in the same call that we are getting all the file info.
If the version summaries api call takes too much time then it will increase the whole file page loading time (this is what is happening on the dataset page with the versions call).
I will make this fix/refactor for the Dataset page, if you want you can wait until I open that PR and follow the same approach.
We should use a custom hook that manages the data, error and loading states an its called from inside the `<FileVersions />` component.

src/sections/Route.enum.ts
```diff
@@ -76,7 +76,7 @@ export enum QueryParamKey {
   PAGE = 'page',
   COLLECTION_ID = 'collectionId',
   TAB = 'tab',
-  FILE_ID = 'fileId',
+  FILE_ID = 'id',
```
g-saracca, line 79: Is there any reason for changing this?

src/sections/file/file-version/FileVersions.tsx
```diff
@@ -0,0 +1,122 @@
...
+  datasetVersionNumber: DatasetNonNumericVersion | DatasetVersionNumber | string | undefined
+}
+
+export function FileVersions({ version, datasetVersionNumber }: FileVersionProps) {
+  const navigate = useNavigate()
+  const { t } = useTranslation('file')
+  const fileId = version?.[0]?.datafileId
+
+  datasetVersionNumber == DatasetNonNumericVersion.DRAFT &&
+    (datasetVersionNumber = DatasetNonNumericVersionSearchParam.DRAFT)
```
g-saracca, line 30: I guess this 👆 was left here unintentionally

src/sections/file/file-version/useFileVersionSummaryDescription.tsx
```diff
@@ -0,0 +1,65 @@
+import {
```
g-saracca, line 1: I think in here we shouldn't use the strings as we receive them from the use case but use our own translations instead as it was implemented on `useDatasetVersionSummaryDescription`.

g-saracca (changes requested):

src/sections/Route.enum.ts
ChengShi-1: It's changed because in file page `http://localhost:8000/spa/files?id=92&datasetVersion=1.0`, QueryParamKey need to be `id` instead of `fileId` to get the file id correctly.

g-saracca: ok, then in `EditFileMetadataWithParams` instead of using a string 'id' could you please use `QueryParamKey.FILE_ID`?

src/sections/file/file-version/FileVersions.tsx
ChengShi-1: Sorry, the code is confused here. I put this line because the `datasetVersionNumber` from the api response will have `:draft`(DatasetNonNumericVersion.DRAFT), but it should be shonw as `DRAFT` (DatasetNonNumericVersionSearchParam.DRAFT).  Is there a better way to deal with the conversion?

src/sections/file/file-version/FileVersions.tsx
```diff
@@ -80,18 +78,16 @@ export function FileVersions({
               <td style={{ verticalAlign: 'middle' }}>
```
g-saracca, line 78: The conditions to render the link is a little bit complex so moving part of that logic outside the rendering could be nice.

You could have constants like this:
```
const isCurrentVersion = fileVersion.datasetVersion === displayVersion;
const isLinkable =
  fileVersion.versionState === DatasetVersionState.RELEASED ||
  fileVersion.versionState === DatasetVersionState.DEACCESSIONED ||
  (fileVersion.versionState === DatasetVersionState.DRAFT && canEditOwnerDataset);
```
And in the render like this:
```
  {isCurrentVersion ? (
    <strong>{fileVersion.datasetVersion}</strong>
  ) : isLinkable ? (
    <Link
      to={`${Route.FILES}?${QueryParamKey.FILE_ID}=${fileId}&${QueryParamKey.DATASET_VERSION}=${fileVersion.datasetVersion}`}
      data-testid={`file-version-link-${fileVersion.datasetVersion}`}
    >
      {fileVersion.datasetVersion}
    </Link>
  ) : (
    <span>{fileVersion.datasetVersion}</span>
  )}
```

g-saracca (changes requested):
Two minor things and we are good to go.

g-saracca (approved):
Approving, good job!! 🎉

ofahimIQSS:
Looks good to me - merging

PR #701, opened 2025-05-16
Fix: avoid showing app loader when refreshing token

g-saracca:
I've addressed an issue where the app loading spinner was unexpectedly appearing during background authentication token refreshes. This was causing the current page state to reset, which could interrupt users who were in the middle of tasks like filling out forms.

This fix should prevent that interruption.

To easily test this, open the Keycloak admin panel, select the Test instance,  and modify the Access Token Lifespan duration to 10 or 20 seconds (it's set to 5 minutes).

Next, open your browser's developer tools and navigate to the network tab. You should observe that a new token is fetched every 10/20 secs. To verify the persistence of form data, navigate to a page with form fields, such as the "Create collection" page. Enter some values into the fields and confirm that these values are retained even after the new token has been obtained.
Please ping me if you need help with the Keycloak admin panel.

No

No

## Additional documentation:
No

### DIFF
--- a/src/router/ProtectedRoute.tsx
+++ b/src/router/ProtectedRoute.tsx
@@ -16,7 +16,8 @@ export const ProtectedRoute = () => {
   const { token, loginInProgress: oidcLoginInProgress, logIn: oidcLogin } = useContext(AuthContext)
   const { user, isLoadingUser } = useSession()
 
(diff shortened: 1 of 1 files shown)


ChengShi-1 (approved):
looks good

ofahimIQSS:
looks good, merging

PR #708, opened 2025-05-22
704 get version summary when tab clicked

### LINKED ISSUE #704: View Dataset Page: Get Version Summary data when Version tab is clicked
Depending on the size of the Dataset, sometimes the version summary info API can be slow.  So to improve performance of the Dataset Page, we need to remove the API call, and execute it only when the user clicks on the Dataset Versions tab.

g-saracca:

> Depending on the size of the Dataset, sometimes the version summary info API can be slow. So to improve performance of the Dataset Page, we need to remove the API call, and execute it only when the user clicks on the Dataset Versions tab.
Also the version summary info was fetched in parallel when getting the dataset details, I removed that.
There was another place where this info was used which is the deaccession dataset modal, now we are only fetching the version summaries info when the modal is opened and only once, meaning that if it was opened before and the internal version number of the dataset remains the same it will use the info that was already fetched before.

- Closes #704

### DIFF
--- a/src/dataset/domain/models/Dataset.ts
+++ b/src/dataset/domain/models/Dataset.ts
@@ -1,7 +1,6 @@
 import { Alert, AlertMessageKey } from '../../../alert/domain/models/Alert'
 import { UpwardHierarchyNode } from '../../../shared/hierarchy/domain/models/UpwardHierarchyNode'
 import { FileDownloadSize } from '../../../files/domain/models/FileMetadata'
-import { DatasetVersionSummaryInfo } from '@/dataset/domain/models/DatasetVersionSummaryInfo'
 
 export enum DatasetLabelSemanticMeaning {
   DATASET = 'dataset',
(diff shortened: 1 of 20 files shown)


ofahimIQSS:
@g-saracca It looks like deaccession is broken here.

Steps to reproduce: Login, create and publish a dataset, then deaccession it. Notice how it fails in the video below. 

https://github.com/user-attachments/assets/b4f3be22-ea93-4673-a728-17b9111f26fa

ofahimIQSS:
@ChengShi-1 / @g-saracca Need a review please 💯

ofahimIQSS:
Adding testing Evidences below:

Screenshot 2025-05-23 at 3 29 32 PM
Screenshot 2025-05-23 at 3 30 04 PM
Screenshot 2025-05-23 at 3 30 54 PM
Screenshot 2025-05-23 at 3 31 39 PM

ChengShi-1 (approved):

PR #718, opened 2025-05-27
Fix Chromatic UI tests

### LINKED ISSUE #706: Chromatic UI tests
Chromatic UI tests and reviews are indicating changes on pull requests even when the touched components haven't been directly modified.

This is likely happening because some of our stories are utilizing dynamic data generated by faker-js instead of relying on static data from our Mother objects. This dynamic data can cause unintentional visual differences between Chromatic builds.

To address this, we should conduct a review of the changes in each story within our Chromatic reports. The goal is to identify any instances where mocked data might be causing these unexpected visual changes and make that data static.

g-saracca:
Chromatic UI tests and reviews are indicating changes on pull requests even when the touched components haven't been directly modified.
This was mostly happening because some of our stories were utilizing dynamic data generated by faker-js instead of relying on static data from our Mother objects. This dynamic data can cause unintentional visual differences between Chromatic builds.

I create some realistic data instead from mother objects to avoid dynamic data and also relied on the `STORYBOOK_CHROMATIC_BUILD` env variable for two components.

I also modify the chromatic workflows to also run when merging against develop so the chromatic baseline is always develop and we have Storybooks build from the latest commit on develop branch.

- Closes #706 

Nothing to test here but to check that the tests have passed.
The work done here will be really reflected in the next Chromatic UI Tests of the upcoming PR's-.

### DIFF
--- a/src/sections/shared/form/ContactForm/ContactCaptcha.tsx
+++ b/src/sections/shared/form/ContactForm/ContactCaptcha.tsx
@@ -3,12 +3,14 @@ import { useTranslation } from 'react-i18next'
 import { Form } from '@iqss/dataverse-design-system'
 import { Validator } from '@/shared/helpers/Validator'
 
+const isChromaticBuild = import.meta.env.STORYBOOK_CHROMATIC_BUILD === 'true'
+
 export function Captcha() {
   const { t } = useTranslation('shared')
   const { control } = useFormContext()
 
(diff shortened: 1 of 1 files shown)


ekraffmiller:
Ok I tested how the seeding is working, and it is running once per story.  So I think that will keep the Faker data deterministic with out needing more changes.

ofahimIQSS:
tests passing - merging.

PR #722, opened 2025-05-28
 Fix search query param and add Types count

### LINKED ISSUE #719: Collection Page - Fix search query param and add Types count
The search query param from the collection page search and the my data page search are being encoded twice.

g-saracca:
Fix a problem were the search query param from the collection page search and the my data page was being encoded twice.
Also adds the count on the Collections, Datasets and Files filter labels in both the collection page and my data page.

Collection page 👇
Screenshot 2025-05-28 at 09 00 00

My Data page 👇

Screenshot 2025-05-28 at 09 00 07

- Closes #719 

Step 1: Run the Development Environment

1. Execute npm i.
2. Navigate with cd packages/design-system && npm i && npm run build.
3. Return with cd ../../.
4. Ensure you have a .env file similar to .env.example
5. Navigate with cd dev-env.
6. Before running next step, be sure have a fresh unstable dataverse image.
7. Start the environment using ./run-env.sh unstable.
8. To verify the environment, visit http://localhost:8000/ and check your local Dataverse installation.

Step 2: Test the search fix and items count

- Check that the search is working properly by checking the results. Perform search from the Homepage, Collection page and Account => My Data section.
- Also, for example, when searching the term "hello world" in the collection page, you will now see in the browser url this `/spa/collections?q=hello+world` instead of this `/spa/collections?q=hello%2520world`.
- Also validate the Collections, Datasets and Files count in both collection page and my data section.

Collection page 👇
Screenshot 2025-05-28 at 09 00 00

My Data page 👇

Screenshot 2025-05-28 at 09 00 07

### DIFF
--- a/tests/component/sections/account/MyDataItemsPanel.spec.tsx
+++ b/tests/component/sections/account/MyDataItemsPanel.spec.tsx
@@ -71,7 +71,7 @@ describe('MyDataItemsPanel', () => {
       cy.findByText(/No results found for user testUserName./).should('exist')
     })
   })
-  describe.only('NoItemsMessage', () => {
+  describe('NoItemsMessage', () => {
     it('renders correct no items message when there are no collection, dataset or files', () => {
       collectionRepository.getMyDataItems = cy.stub().resolves(emptyItemsWithCount)
 


ChengShi-1:
I test it and find the order of the result is different between JSF and SPA. 
For example, search "blob-1", the JSF will return "blob-1" first in the list, SPA will return "blob" first. It seems like a back-end issue here?

https://github.com/user-attachments/assets/0db1d5dd-1c18-44cb-aa1d-6857323da2b1

g-saracca:
> I test it and find the order of the result is different between JSF and SPA. For example, search "blob-1", the JSF will return "blob-1" first in the list, SPA will return "blob" first. It seems like a back-end issue here?
> 
>  Screen.Recording.2025-05-28.at.10.18.27.AM.mov

Yep, I can see the same, but it sounds like an API/Backend issue as we are rendering the items in the same order as we receive them from the API response 👍🏼 .
So I believe this should not be a blocker for this PR.

ChengShi-1 (approved):
Great job! looks good

PR #732, opened 2025-05-30
Fix CollectionItemsPanel flaky test

### LINKED ISSUE #724: Fix Flaky e2e Test: CollectionItemsPanel.spec.ts
Sometime CollectionItemsPanel.spec.ts fails in github action, for example: https://github.com/IQSS/dataverse-frontend/actions/runs/15307334323/job/43063246624.  It could be a timing issue for datasets being created and destroyed in the root collection.  
A possible fix would be to add the datasets to be tested in a sub-collection, rather than the root collection.

ekraffmiller:
CollectionItemsPanel.spec.ts was failing intermittently on Github actions.  I was able to reproduce it consistently by using this command, which runs the test 5 times in a row:

`npx cypress run --env burn=5 --spec tests/e2e-integration/e2e/sections/collection/CollectionItemsPanel.spec.ts
`

After making the fix, this test passes.

- Closes #724 

Try running the batch test locally, and check that the github test also succeeds.


ChengShi-1 (approved):
looks good

PR #742, opened 2025-06-17
My Data - search for username when component renders

### LINKED ISSUE #738: Account Page My Data section: execute user search on page load
When the currently logged in user is a super user, the user search input field is displayed with the user's name by default.  The search is done when the user triggers a submit from the search input, but it should also be done when the page is first loaded.

Image

ekraffmiller:

- Closes #738 

I also updated the parameter types for GetMyDataCollectionItems, to match js-dataverse
Note that there is an issue with the MyData API call (https://github.com/IQSS/dataverse/issues/11519), where it uses the jsf user session.  So it's best to test this without also having a JSF tab open in the browser. 

For the Chromatic tests: The deposit date is different because it is set in the component, not by mock data.  I'm not sure what is happening with the FileInfoCell story, because it uses the same static static seed, maybe because the stories were executed in a different order, I'm not sure.

1. Login as a superuser, for example dataverseAdmin. 
2. Go To Accounts -> My Data
3. Check that the API call made when My Data is displayed includes the username, 'dataverseAdmin'

### DIFF
--- a/src/sections/account/my-data-section/MyDataItemsPanel.tsx
+++ b/src/sections/account/my-data-section/MyDataItemsPanel.tsx
@@ -143,7 +143,8 @@ export const MyDataItemsPanel = ({ collectionRepository }: MyDataItemsPanelProps
       [CollectionItemType.COLLECTION, CollectionItemType.DATASET, CollectionItemType.FILE],
       roleIds,
       AllPublicationStatuses,
-      searchValue === '' ? undefined : searchValue
+      searchValue === '' ? undefined : searchValue,
+      currentSearchCriteria.otherUserName
     )
 
     const totalItemsCount = await loadMore(resetPaginationInfo, newCollectionSearchCriteria, true)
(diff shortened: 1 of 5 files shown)


src/sections/account/my-data-section/MyDataItemsPanel.tsx
```diff
@@ -57,7 +57,8 @@ export const MyDataItemsPanel = ({ collectionRepository }: MyDataItemsPanelProps
       [CollectionItemType.COLLECTION, CollectionItemType.DATASET],
       roleIds,
       AllPublicationStatuses,
-      undefined
+      undefined,
+      user?.superuser ? user.identifier : undefined
     )
   )
```
g-saracca, line 63: I've noticed that the user identifier isn't being included in the API calls when the type, role, or publication status is changed. I believe it should always be sent if there is a value entered, right?

So besides of adding the super user identifier by default in the SearchCriteria you could probably add it in:
`handlePublicationStatusChange` => `newMyDataSearchCriteria`
`handleItemsTypeChange` => `newMyDataSearchCriteria`
`handleRoleChange` => `newMyDataSearchCriteria`
And probably also in `handleSearchSubmit`?

g-saracca (approved):

ofahimIQSS:
no issues found during test - merging.

PR #750, opened 2025-06-26
Account Page My Data: Integrate roles use case 

### LINKED ISSUE #685: Account Page -> MyData section: integrate roles use case
Create a frontend use case to get user roles, and integrate it with the My Data Section for filtering collection items. 
 Depends on this js-dataverse issue: https://github.com/IQSS/dataverse-client-javascript/issues/300

ekraffmiller:

- Closes #685 

I made some changes to the featured items code, to be compatible with the latest changes in js-dataverse.

Login as different types of users, check that the list of Roles in the My Data filter panel are correct.
For dataverseAdmin (superUser)  all roles should be displayed
For a new user, with no assigned roles, all roles should be displayed
When a user is assigned a role, only the list of assigned roles should be displayed
Check that the list of roles matches the roles returned from the 'retrieve' API in the developer's console.

no


tests/component/sections/account/MyDataItemsPanel.spec.tsx
```diff
@@ -67,22 +70,50 @@ describe('MyDataItemsPanel', () => {
...
+    cy.mountAuthenticated(
+      <MyDataItemsPanel
+        roleRepository={roleRepository}
+        collectionRepository={collectionRepository}
+      />
+    )
+
+    cy.findByRole('alert').should('exist').should('contain.text', 'some roles error')
+  })
   describe('User Search', () => {
```
ChengShi-1, line 98: Could you give an empty line before describe() block for formatting purpose

tests/component/sections/account/MyDataItemsPanel.spec.tsx
```diff
@@ -67,22 +70,50 @@ describe('MyDataItemsPanel', () => {
...
+    cy.mountAuthenticated(
+      <MyDataItemsPanel
+        roleRepository={roleRepository}
+        collectionRepository={collectionRepository}
+      />
+    )
 
     cy.findByTestId('collection-items-list-infinite-scroll-skeleton').should('exist')
   })
+  it('renders the error message when there is an error', () => {
```
ChengShi-1, line 86: Could you give an empty line before it() block for formatting purpose

src/sections/account/my-data-section/useSelectableRoles.tsx
```diff
@@ -0,0 +1,30 @@
...
+  const [isLoading, setIsLoading] = useState<boolean>(true)
+  const [error, setError] = useState<string | null>(null)
+
+  useEffect(() => {
+    const fetchRoles = async () => {
+      try {
+        const fetchedRoles = await roleRepository.getUserSelectableRoles()
+        setRoles(fetchedRoles)
+      } catch (err) {
+        const errorMessage =
```
ChengShi-1, line 16: `setRoles([])` // clear roles to reflect failure

src/sections/account/my-data-section/useSelectableRoles.tsx
```diff
@@ -0,0 +1,30 @@
...
+export const useSelectableRoles = (roleRepository: RoleRepository) => {
+  const [roles, setRoles] = useState<Role[]>([])
+  const [isLoading, setIsLoading] = useState<boolean>(true)
+  const [error, setError] = useState<string | null>(null)
+
+  useEffect(() => {
+    const fetchRoles = async () => {
+      try {
+        const fetchedRoles = await roleRepository.getUserSelectableRoles()
+        setRoles(fetchedRoles)
```
ChengShi-1, line 14: `setError(null)`  // clear previous error

src/sections/account/my-data-section/useSelectableRoles.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { Role } from '@/roles/domain/models/Role'
+
+export const useSelectableRoles = (roleRepository: RoleRepository) => {
+  const [roles, setRoles] = useState<Role[]>([])
+  const [isLoading, setIsLoading] = useState<boolean>(true)
+  const [error, setError] = useState<string | null>(null)
+
+  useEffect(() => {
+    const fetchRoles = async () => {
+      try {
```
ChengShi-1, line 12: setIsLoading(true)
setRoles([])

 // reset before fetch

src/sections/account/my-data-section/useSelectableRoles.tsx
```diff
@@ -0,0 +1,30 @@
+import { useState, useEffect } from 'react'
+import { RoleRepository } from '@/roles/domain/repositories/RoleRepository'
+import { Role } from '@/roles/domain/models/Role'
+
+export const useSelectableRoles = (roleRepository: RoleRepository) => {
+  const [roles, setRoles] = useState<Role[]>([])
+  const [isLoading, setIsLoading] = useState<boolean>(true)
+  const [error, setError] = useState<string | null>(null)
+
```
ChengShi-1, line 9: When the API call fails, the roles state remains as it was before, possibly from a previous successful fetch. This means:
- It will show stale or outdated data.
- The roles array is not cleared on error, so the page may misleadingly show data even though the fetch failed.

Also, we could have a test for this hook here

ChengShi-1 (commented):
Good! I left some comments, please check them. 
Also,
- `package.json` could be updated. 
- The PR for featured items got merged, so this PR could merged with `dev` branch again.

src/sections/account/my-data-section/useSelectableRoles.tsx
ekraffmiller: I think that is a good idea in general about cleaning up stale data, but in this case I don't think it's needed.
The state is initialized every time the component (MyDataItemsPanel) is mounted, and is only called once when the component is mounted.  If it fails, it wouldn't be called again until the next time the component is mounted, and at that time the state will be initialized again.  So to keep the logic simple, I think it's ok to not update the state.  What do you think?
I added a test to MyDataItemsPanel, to  test that the correct roles from the hook are being displayed 👍

ChengShi-1 (approved):
Everything looks good!

PR #776, opened 2025-07-18
Dataset Page: Guest User can't get file info for a deaccessioned dataset 

### LINKED ISSUE #773: Prevent guests of a deaccessioned dataset from getting file's info
Based on changes https://github.com/IQSS/dataverse/pull/11636 here, `download/count` and `downloadsize` will return 403 for user: guests, the `getByPersistentId` should be refactored to adapt the change

If `user:guest` and the dataset is deaccessioned:

- dataset is ok to show basic info, versions and deaccessioned reason
- NO File info could be shown, cannot direct to file page as well
- Hide share button
- Hide metric(prevent fetching `download/count`

ChengShi-1:
download/count and downloadsize will return 404 for user: guests, the getByPersistentId should be refactored to adapt the change

If user:guest and the dataset is deaccessioned:

dataset is ok to show basic info, versions and deaccessioned reason
NO File info could be shown, cannot direct to file page as well
Hide share button
Hide metric(prevent fetching download/count

- Closes #773 

⚠️ waiting for https://github.com/IQSS/dataverse/pull/11702

As a guest user, in a dataset, check if the dataset metric component and the Share button is hidden. As a dataset owner, check if the dataset metric component and the Share button is shown.

image


src/sections/dataset/dataset-action-buttons/DatasetActionButtons.tsx
```diff
@@ -54,7 +58,7 @@ export function DatasetActionButtons({
           contactRepository={contactRepository}
         />
 
-        <ShareDatasetButton />
+        {(!isCurrentVersionDeaccessioned || canUpdateDataset) && <ShareDatasetButton />}
```
g-saracca, line 61: Is this logic replicating JSF condition?

src/sections/dataset/Dataset.tsx
```diff
@@ -144,10 +144,13 @@ export function Dataset({
...
-                datasetRepository={datasetRepository}
-                datasetId={dataset.persistentId}
-              />
+              {(!isCurrentVersionDeaccessioned || canUpdateDataset) && (
+                <DatasetMetrics
+                  data-testid="dataset-metrics"
+                  datasetRepository={datasetRepository}
+                  datasetId={dataset.persistentId}
+                />
+              )}
```
g-saracca, line 158: Is this logic replicating JSF condition?

tests/e2e-integration/integration/datasets/DatasetJSDataverseRepository.spec.ts
```diff
@@ -378,27 +372,26 @@ describe('Dataset JSDataverse Repository', () => {
...
+    await datasetRepository.getByPersistentId(datasetResponse.persistentId).then((dataset) => {
+      if (!dataset) {
+        throw new Error('Dataset not found')
+      }
+      expect(dataset.version.publishingStatus).to.equal(DatasetPublishingStatus.DEACCESSIONED)
+      expect(dataset.fileDownloadSizes).to.deep.equal([
+        new FileDownloadSize(0, FileSizeUnit.BYTES, FileDownloadMode.ORIGINAL),
+        new FileDownloadSize(0, FileSizeUnit.BYTES, FileDownloadMode.ARCHIVAL)
+      ])
+    })
```
g-saracca, line 418: Why are we deleting the test 'gets the dataset by persistentId when is locked'?

src/dataset/infrastructure/repositories/DatasetJSDataverseRepository.ts
```diff
@@ -201,17 +201,20 @@ export class DatasetJSDataverseRepository implements DatasetRepository {
...
             return datasetDetails
           })
-          .catch(() => {
-            return datasetDetails
+          .catch((error: ReadError) => {
+            if (error.message.includes('404')) {
+              // If the server returns NOT_FOUND when deaccessioned info isn't included, ignore and continue without sizes.
+              return datasetDetails
+            }
+            throw error
```
g-saracca: I dont understand this, why this checking for 404? Is the description of this PR outdated?
In steves PR I see this comment "guest user should see deaccessioned file count and download size"

g-saracca: Also you could use `JSDataverseReadErrorHandler` and don't cast the error `error: ReadError`, as you could have other many types of errors. Try something like this:
```
} catch (err: ReadError | unknown) {
        if (err instanceof ReadError) {
          const error = new JSDataverseReadErrorHandler(err)
          const statusCode = error.getStatusCode()
          if (statusCode === 404) return datasetDetails
        } else if (err instanceof Error && err.message.includes('404')) {
          return datasetDetails
        } else {
          throw error
        }
}
```

g-saracca (changes requested):
Hi @ChengShi-1 , some comments here 👍🏼

src/dataset/infrastructure/repositories/DatasetJSDataverseRepository.ts
ChengShi-1: @g-saracca Sorry, the description said it should return 403, but it actually it returns 404.

- For guest users with includeDeaccessioned=false: Returns 404 Not Found
- For guest users with includeDeaccessioned=true: Returns 200 OK with download size.

Also, alternatively, if the dataset is deaccessioned and canEditDataset is false, skip calling the downloadSize endpoints and keep the defaults, avoiding the error path entirely. Do you think it will be better?

tests/e2e-integration/integration/datasets/DatasetJSDataverseRepository.spec.ts
ChengShi-1: I wrongly overwrite this test while merging conflict :( I'll recover this

src/sections/dataset/Dataset.tsx
ChengShi-1: Yes, the deaccessioned dataset page will hide metric and share button to guest users

src/dataset/infrastructure/repositories/DatasetJSDataverseRepository.ts
g-saracca: Sorry I still don't understand. 
If you are a guest user and do NOT request to include deaccessioned version you will receive a 404, but you will receive a 200 ok being a guest user requesting a deaccessioned version? how can be that way?
I'm lost with this thing sorry 🙏🏼

ChengShi-1: Yes, guest user could still access files' downloadSize info for deaccessioned dataset, which will work for a deaccessioned [tombstone](https://guides.dataverse.org/en/6.6/user/dataset-management.html#dataset-deaccession) 

> Important Note: A tombstone landing page with the basic citation metadata will always be accessible to the public if they use the persistent URL (Handle or DOI) provided in the citation for that dataset. Users will not be able to see any of the files or additional metadata that were previously available prior to deaccession.

ChengShi-1: For the backend, the issue was `includeDeaccessioned=true, user: guest` returns 404, so they fixed, now it returns 200.

In the UI, if we don't skip the 404, the UI will be 404 error page.
The `downloadSize` is a part of `getByPersistentId` in `datasetJSDataverseRepository`, and `fileDownloadSizes` is a field of the Dataset model. In order to prevent the showing of 404 error page, I catches 404 here. 

You should not be sorry, this is truly confusing, I am not sure if I did it right. Thanks for helping!

g-saracca: Ok I think I understand now, try to use the suggestion I made for catching the error in a more robust way.

g-saracca (approved):

ofahimIQSS:
image

Looks good to me, merging.

PR #811, opened 2025-08-21
Fix - "You can only drop one file to replace" warning on upload mode

### LINKED ISSUE #810: Bug - "You can only drop one file to replace" warning on upload mode
We have a client side validation to avoid user dropping more than one file while replacing a file. But is happening on Upload mode.

g-saracca:
It's a one liner. It corrects an error whereby, when loading files into a Dataset and dropping more than one file, a validation occurred that should only have occurred during replacing a file.

- Closes #810 

Create a dataset and go to the upload files page. Drop two files at the same time, the "You can only drop one file to replace" toast error should not appear.

No

No

## Additional documentation:
No


ChengShi-1 (approved):

ofahimIQSS:
looks good, merging.

PR #833, opened 2025-09-17
recommend latest LTS version of node

pdurbin:

v19 is EOL according to https://nodejs.org/en/about/previous-releases

None.

I picked v10 for npm because it's what came along with node when I installed the LTS of node (v22).

I noticed that in GitHub Actions we use v18, which is also EOL, but I'd suggest a different pull request for that:

```
% grep -r node-version .github/workflows 
.github/workflows/lint.yml:          node-version: 18
.github/workflows/accessibility.yml:          node-version: 18
.github/workflows/test.yml:          node-version: 18
.github/workflows/test.yml:          node-version: 18
.github/workflows/deploy.yml:          node-version: 18
.github/workflows/deploy-beta-testing.yml:          node-version: 18
```

@g-saracca noticed that the Docker environment is using v19:

```
% grep FROM dev.Dockerfile
FROM node:19.6.1-alpine AS build_image
FROM node:19.6.1-alpine 
```

### DIFF
--- a/DEVELOPER_GUIDE.md
+++ b/DEVELOPER_GUIDE.md
@@ -73,7 +73,7 @@ _To get a local copy up and running follow these simple example steps._
 
 [![DockerDesktop][_shield_docker]][_uses_docker_url]
 
-1. **Node &amp; NPM**: `node >= v16` and `npm >= v8`. Recommended versions for this project are `node v19` and `npm v9`.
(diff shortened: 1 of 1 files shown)


g-saracca:
Hi @pdurbin, do you think we can close this PR? I already tackle this and lots of other dependencies vulnerabilities on https://github.com/IQSS/dataverse-frontend/pull/840 that I've just opened.

pdurbin:
@g-saracca yes! It looks like you have my change covered in that PR. Thanks! Closing!

PR #838, opened 2025-09-23
Enhance default username while registering with academic account

### LINKED ISSUE #837: Create Account: username provided by Identity Provider (IdP) can be incompatible with Dataverse usernames while creating account
This feature request aims to request the ability to change username while registering an account with university email. It happened in beta.dataverse.org, while I was in the process for registering and taking a look in frontend app. However, I was not able to complete registration as the username had a constant value provided by my university which included this char '@' and I was not able to change it.

Image

Image

superpegaso2703:
This PR tries to handle default username if it includes a '@' character, which is not allowed.
- Closes #837
## Suggestions on how to test this (to reproduce without the code of this pull request):
1. Once the dataverse frontend repository has been cloned, navigate to dev-env dir.
2. Modify /dataverse-frontend/dev-env/keycloak/test-realm.json and set Line 443 to this value ```"username": "dataverse-curator@mailinator.com"```
3. Execute ```./run-env.sh latest```.
4. Proceed to navigate to http://localhost:8000/spa and click login button.
5. Introduce it's email and password in test realm.
6. After successfull login, the browser will redirect to http://localhost:8000/spa/sign-up?validTokenButNotLinkedAccount=true.
7. After completing required data and trying to login, message of invalid username will be received.
8. Photos are included in additional documentation of this pull request.
Yes, the error which displayed invalid username should not happen again.
Yes, for example an enhancement of login process due to fix of username including special and not allowed characters.
## Additional documentation:
image
Screenshot From 2025-09-28 19-19-07


superpegaso2703:
Modified a test, going to include output here.
```npx cypress run --spec tests/e2e-integration/e2e/sections/auth/Login.spec.tsx

====================================================================================================

  (Run Starting)

  ┌────────────────────────────────────────────────────────────────────────────────────────────────┐
  │ Cypress:        12.14.0                                                                        │
  │ Browser:        Electron 106 (headless)                                                        │
  │ Node Version:   v20.19.2 (/usr/bin/node)                                                       │
  │ Specs:          1 found (Login.spec.tsx)                                                       │
  │ Searched:       tests/e2e-integration/e2e/sections/auth/Login.spec.tsx                         │
  └────────────────────────────────────────────────────────────────────────────────────────────────┘


────────────────────────────────────────────────────────────────────────────────────────────────────
                                                                                                    
  Running:  Login.spec.tsx                                                                  (1 of 1)


  Login
    ✓ successfully log in with a user that exists in dataverse and not in the OIDC provider (5755ms)
    ✓ successfully log in and finish the sign up with a user that exists in the OIDC provider and not in dataverse (7451ms)
    ✓ successfully log in and finish the sign up with a user that exists in the OIDC provider and not in dataverse with a user who has email as username (5254ms)


  3 passing (19s)


  (Results)

  ┌────────────────────────────────────────────────────────────────────────────────────────────────┐
  │ Tests:        3                                                                                │
  │ Passing:      3                                                                                │
  │ Failing:      0                                                                                │
  │ Pending:      0                                                                                │
  │ Skipped:      0                                                                                │
  │ Screenshots:  0                                                                                │
  │ Video:        false                                                                            │
  │ Duration:     18 seconds                                                                       │
  │ Spec Ran:     Login.spec.tsx                                                                   │
  └────────────────────────────────────────────────────────────────────────────────────────────────┘


====================================================================================================

  (Run Finished)


       Spec                                              Tests  Passing  Failing  Pending  Skipped  
  ┌────────────────────────────────────────────────────────────────────────────────────────────────┐
  │ ✔  Login.spec.tsx                           00:18        3        3        -        -        - │
  └────────────────────────────────────────────────────────────────────────────────────────────────┘
    ✔  All specs passed!                        00:18        3        3        -        -        -  
```

superpegaso2703:
I think if the tests fail here in github it could be due to the order of being uploaded. However I registered using the updated ```Login.spec.tsx``` and logged in manually after it.
image

g-saracca:
Hi @superpegaso2703 , thanks for your contribution! 🎉 I can see your branch is several commits behind develop, where we've fixed several things, including dependencies upgrades (node 22 required now) and tests fixes (including e2e).
If you could update your branch would be great or if you want I can do it for you.

pdurbin:
@g-saracca if you want, we could check this box at https://github.com/IQSS/dataverse-frontend/settings

Screenshot 2025-10-07 at 7 35 38 AM

I discovered this checkbox recently. You may have seen me talk about it in Slack. I also blogged about it. 😄 http://blog.greptilian.com/2025/09/26/a-button-on-github-to-merge-the-latest-into-your-pull-request/

g-saracca:
> https://github.com/IQSS/dataverse-frontend/settings

Hey @pdurbin, yeah I remember, sounds handy!  although I'm not sure it will work with forks, but let's try it 💯

g-saracca:
I think I’ve identified a blocker about handling this from the frontend side.
The current fix doesn’t actually impact how the user is stored in the database because of the approach followed in the API side. The default username is only meant to be displayed in the form;  we’re not sending those values to the `/register` endpoint [if they are present in the token data](https://github.com/IQSS/dataverse-frontend/blob/905d6107bfb2253cf5ec74e12486c799e91cfaa2/src/sections/sign-up/valid-token-not-linked-account-form/ValidTokenNotLinkedAccountFormHelper.ts#L55), otherwise the endpoint will fail when noticing the username sent is different from the one present in token data.

That’s also why this banner message appears:

> “Some fields in this form are pre-filled with information from your identity provider and cannot be changed at this stage. This ensures consistency and security. Please complete the remaining fields to finish your registration.”

Anyways, I gave it a try:

I've created this user 👇
Screenshot 2025-10-07 at 09 12 25

And sent the username anyway (by commenting out the first line below this [comment](https://github.com/IQSS/dataverse-frontend/blob/905d6107bfb2253cf5ec74e12486c799e91cfaa2/src/sections/sign-up/valid-token-not-linked-account-form/ValidTokenNotLinkedAccountFormHelper.ts#L55) and sending the username in the DTO), see result.

Screenshot 2025-10-07 at 09 21 07

 My conclusion is that this should be handled on the Keycloak/backend side, @GPortas sounds good?

Sorry for not noticing this earlier 🙏🏼

superpegaso2703:
@g-saracca I think it's usually a good practice to validate in frontend and in backend too, so frontend and backend are  synchronized, in HTB CWEE path I remember exploiting an issue called validation logic disparity in which It affected frontend and backend discrepancies.

g-saracca:
> @g-saracca I think it's usually a good practice to validate in frontend and in backend too, so frontend and backend are syncrhonized, in HTB CWEE path I remember exploiting an issue called validation logic disparity in which It affected frontend and backend discrepancies.

I totally agree, we also perform many validations both on the frontend and backend to stay in sync and avoid discrepancies like the one you mentioned.
However, in this particular case, even though the fix itself is simple, it wouldn’t work properly with the current logic of the `/register` endpoint since this will fail if we send a different username.
In my opinion, it would make more sense for the backend to handle the username modification so it remains consistent with Dataverse’s rules and we keep that strong security in the registration endpoint, anyways open to ideas 👍🏼 

What we could do, I think, is changing the frontend username rules validations to allow "@" in the username?

```
  const userNameRules = {
    required: isUsernameRequired ? t('fields.username.required') : false, // <= <= We only require the username if not present in token data
    validate: (value: string) => {
      if (!Validator.isValidUsername(value)) { // <= <= <= THIS PART IS SHOWING THE ERROR
        return t('fields.username.invalid.username')
      }
      return true
    },
    maxLength: {
      value: 255,
      message: t('fields.username.invalid.maxLength', {
        maxLength: 255
      })
    }
  }
```

GPortas:
Yes, as @g-saracca mentioned, in this case we need to delegate the registration logic to the backend or Keycloak. The final username should be determined by the backend or Keycloak based on the exact value coming from the OIDC token claim.

At this point, I can think of two possible ways to address this issue:

1. Custom validation in Keycloak: Implement a custom Keycloak mechanism that prevents users who log in with an email-formatted username from being required to update it before continuing to the SPA.
2. Backend-side handling: Allow the backend to accept email-type usernames and apply any necessary conversions transparently to the user to generate the final username (we could even consider whether it makes sense to allow email-style usernames directly).

I’d probably go with the first option, since our goal with introducing Keycloak is to fully delegate authentication responsibilities to it, rather than keeping them within the Dataverse backend service.

@superpegaso2703 Keeping frontend and backend validations consistent is definitely important to avoid discrepancies. That said, in this specific case the frontend shouldn’t be responsible for knowing or enforcing this particular constraint. Since the rules for username handling come directly from the backend or Keycloak (and could evolve over time), duplicating that logic in the frontend would increase coupling, maintenance overhead, and could result in the validation logic disparity issue you mentioned.

From a separation of concerns perspective, the frontend should focus only on UI and basic UX-related validations (format, required fields, etc.), while the backend or Keycloak (ideally Keycloak) serves as the single source of truth for authentication-related rules like this one.

GPortas:
@superpegaso2703 As Germán mentioned, we’re sorry for not realizing this earlier and for you having continued with development and testing. We’ll discuss this in more detail in order to reach a consensus on which direction to take to fix your issue.

PR #858, opened 2025-10-06
ignore onClick logOut when calculating code coverage

pdurbin:

Ignores the following line when calculating code coverage:

Screenshot 2025-10-06 at 2 42 26 PM

None. Relates to:

- #866

If we can write a test to exercise logOut, that would probably be better.

Interact with the header.

No.

No.

## Additional documentation:

None.

### DIFF
--- a/src/sections/layout/header/LoggedInHeaderActions.tsx
+++ b/src/sections/layout/header/LoggedInHeaderActions.tsx
@@ -69,7 +69,10 @@ export const LoggedInHeaderActions = ({
           {t('navigation.apiToken')}
         </Navbar.Dropdown.Item>
 
-        <Navbar.Dropdown.Item href="#" onClick={() => logOut()} data-testid="oidc-logout">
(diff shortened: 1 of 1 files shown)


pdurbin:
> If we can write a test to exercise logOut, that would probably be better.

I spoke with @g-saracca and he's pretty sure we can write such a test. If we come up with a better way to cover this line instead of ignoring it, please feel free to close this PR! 😅

g-saracca:
@pdurbin PR to improve coverage ready [here](https://github.com/IQSS/dataverse-frontend/pull/869).
[This is the line](https://github.com/IQSS/dataverse-frontend/blob/9e0401bb0c39e98bf8a634d9169122ab2e910fab/tests/component/sections/layout/header/LoggedInHeaderActions.spec.tsx#L98) covering this logout case for example.

pdurbin:
@g-saracca fantastic! Closing!

PR #874, opened 2025-10-15
Dataset Edit Terms Integration

### LINKED ISSUE #824: Dataset Edit Terms Integration
This feature will mainly focus on  “Dataset Terms”, “Restricted Files + Terms of Access” UI implementation. 

Based on the spike issue https://github.com/IQSS/dataverse-frontend/issues/799, edit terms page includes three main update actions: “Dataset Terms”, “Restricted Files + Terms of Access”, and “Guestbook”(The Guestbook section in this page won’t be implemented for now) , we would split these sections into separate tabs, each with its own save button and to consider renaming the “Edit Terms” option in the Edit Dataset dropdown, or even a completely different than this one.

The spike issue and 2025.Q3 proposal  

JSF > Edit Terms page 
Image

[Demo on storybook](https://68a5eb16df09a430a6e757ee-uavlglbqwm.chromatic.com/iframe.html?id=pages-editdatasetterms--dataset-terms-tab&viewMode=story)

Image

ChengShi-1:
This feature will mainly focus on “Dataset Terms”, “Restricted Files + Terms of Access” UI implementation.(Guestbook is not implemented for now)

- Closes #824 

- make edit terms to be an independent route, which acts differently from JSF
- for all versions of dataset, the terms information should be the same, and always update to the latest version
- for mobile viewport of this page, it should become accordion(as the same as JSF) instead of tabs, which looks friendly to mobile users
- In Dataset page, terms of access should be able to take HTML snippets

- A edit button in dataset page, naming changed to Edit Terms and Guestbook suggested from Julian to better guide users
image
- License tab - should be able to select from some available standard licenses or custom terms
image
- Restricted Files + Terms of Access Tab - should be able to enable/disable request access, also Terms of Access for Restricted Files would be required if disable request access
image
- mobile viewport
image

yes, should be added later

yes, should be added later


public/locales/en/dataset.json
```diff
@@ -1,16 +1,19 @@
...
     "description": "The following custom terms have been defined for this dataset."
   },
   "license": {
     "title": "License/Data Use Agreement",
-    "altTextPrefix": "License image for "
+    "altTextPrefix": "License image for ",
+    "description": "This dataset will be published under the terms specified below. Our Community Norms as well as good scientific practices expect that proper credit is given via citation.",
+    "moreInfo": "More information"
   },
+  "loading": "Loading...",
```
ChengShi-1: This loading should be removed

ekraffmiller:
Since I know you are on vacation soon, I'm adding comments of little things as I see them :) 
Can we add more space here between the dropdown and the license image and link?

Screenshot 2025-11-13 at 4 02 14 PM

ekraffmiller:
The cancel buttons on the tabs are resetting the values, but I think they also need to redirect the user to the View Dataset page, since they are exiting out of edit mode. (If they are leaving the page, do the values need to be reset?)


Screenshot 2025-11-13 at 4 15 09 PM

ekraffmiller:
In JSF, if the Dataset doesn't have any restricted files, then the Request Access and Terms of Access for Restricted Files is disabled. It doesn't do any harm to keep them enabled, but it would be nice to match the JSF

Screenshot 2025-11-13 at 4 41 16 PM

ChengShi-1:
> if the Dataset doesn't have any restricted files, then the Request Access and Terms of Access for Restricted Files is disabled.

Hi @ekraffmiller , I check the JSF again

- If there is no file at all, or no restricted file, `Enable Request Access` is disabled.
- If one of the file is restricted, `Enable Request Access` should be enabled, no matter it set `Enable request access` to true or false when the file is restricted in this modal
image

So, the goal is to detect whether any file is restricted. Right now, the only way I can find is to query each file via the file repository use cases, which could be inefficient. The dataset repository doesn’t expose a use case to check for the existence of restricted files. Is there a better pattern for identifying this, or should we request an API extension?

ekraffmiller:
> So, the goal is to detect whether any file is restricted. Right now, the only way I can find is to query each file via the file repository use cases, which could be inefficient. The dataset repository doesn’t expose a use case to check for the existence of restricted files. Is there a better pattern for identifying this, or should we request an API extension?

Thanks @ChengShi-1 , that's a good point!  We don't want to query each file, so I think an API extension is a good idea.  We can leave the Edit Terms behavior as is, since it's not doing any harm.

src/sections/dataset/dataset-action-buttons/edit-dataset-menu/EditDatasetMenu.tsx
```diff
@@ -57,8 +57,16 @@ export function EditDatasetMenu({ dataset, datasetRepository }: EditDatasetMenuP
       return
     }
     if (eventKey === EditDatasetMenuItems.TERMS) {
-      showModal()
-      return
+      const searchParams = new URLSearchParams()
```
ekraffmiller: The persistentId and version searchParams are set above, in the beginning of handleOnSelect(), so we don't need to set them here.

src/sections/edit-dataset-terms/dataset-terms-tab/DatasetTermsTab.tsx
```diff
@@ -0,0 +1,294 @@
...
+      .map((license) => ({
+        value: license.id.toString(),
+        label: license.name,
+        uri: license.uri,
+        iconUri: license.iconUri || '',
+        isDefault: license.isDefault
+      }))
+
+    dynamicOptions.push({
+      value: 'CUSTOM',
```
ekraffmiller: since this hard-coded 'CUSTOM' string is used in multiple places, would be good to replace it with a const.

src/sections/edit-dataset-terms/edit-license-and-terms/EditLicenseAndTerms.tsx
```diff
@@ -0,0 +1,294 @@
...
+      rows: 4,
+      rules:
+        fieldName === 'termsOfUse' && isCustomTerms
+          ? { required: t('editTerms.datasetTerms.customTermsRequired') }
+          : {}
+    }))
+  }, [initialCustomTerms, isCustomTerms, t])
+
+  const onSubmit = (data: DatasetTermsFormData) => {
+    if (!dataset) return
```
ekraffmiller, line 168: if the onSubmit is successful, it should also navigate to the view version of the Dataset page.

src/sections/edit-dataset-terms/EditDatasetTerms.tsx
ekraffmiller, line 1: The naming of the child components within each tab is a little confusing (although I see the logic of it because the component names match the tab names).  But I think it would be clearer to rename the tab components so that they reflect that they are edit components.  I was confused about 'DatasetTermsTab' because there is another component DatasetTerms, for displaying the terms in view mode.

For example,
DatasetTermsTab -> EditLicenseAndTerms
RestrictedFilesTab -> EditTermsOfAccess
GuestBookTab -> EditGuestBook

src/stories/edit-dataset-terms/EditDatasetTerms.stories.tsx
```diff
@@ -0,0 +1,65 @@
...
+    <EditDatasetTerms
+      defaultActiveTabKey={EditDatasetTermsHelper.EDIT_DATASET_TERMS_TABS_KEYS.guestBook}
+      licenseRepository={licenseRepository}
+      datasetRepository={datasetRepository}
+    />
+  )
+}
+
+export const AllTabs: Story = {
+  render: () => (
```
ekraffmiller: What is AllTabs meant to show?  It's currently the same as 'Default'

tests/component/dataset/domain/models/TermsOfUseMother.ts
```diff
@@ -99,8 +71,8 @@ export class TermsOfUseMother {
 
   static createRealistic(props?: Partial<DatasetTermsOfUse>): DatasetTermsOfUse {
     const defaultTerms: DatasetTermsOfUse = {
-      termsOfAccess: TermsOfAccessMother.createRealistic(),
```
ekraffmiller, line 102: Why were the createRealistic() functions removed?  It works with random values in create(), but most of our stories use the realistic versions of mock data, so the Story looks more 'normal'

ekraffmiller (commented):
Hi @ChengShi-1, it looks really good!  See comments about suggested changes.
Sorry it took so long for this review, please don't worry about it till you are back from vacation! :)

ekraffmiller (commented):
Hi @ChengShi-1, all your changes look great, I just had another thought about the UI when testing this.
Currently, a user can make an edit on one tab, then click over to another tab without saving changes.  I could see where a user would be confused, and think that 'Save Changes' means to save the changes made in all tabs within Edit Dataset Terms page.
If the user goes to a second tab without saving the first tab, then the changes will be lost.
I think the best solution for the user experience is to update "Save Changes" so it saves the data from all the tabs, but I realize that might be a big change.  An alternate approach would be to warn the user if they click on another tab while unsaved changes exist on the current tab.  We do something similar in the UploadFiles page - if the user attempts to navigate away from the page, we show a popup warning them, and giving an option to "stay on the page".
What do you think? Does one solution seem easier to implement than the other?

ChengShi-1:
@ekraffmiller I agree the button can be confusing, and acts different from JSF. Since each tab submits separately in its own API and use case now, adding an unsaved-changes warning is simpler. I can add that warning now, and if we want a single save across tabs later, I can tackle that next.

ekraffmiller:
> @ekraffmiller I agree the button can be confusing, and acts different from JSF. Since each tab submits separately in its own API and use case now, adding an unsaved-changes warning is simpler. I can add that warning now, and if we want a single save across tabs later, I can tackle that next.

sounds good, thanks!

ChengShi-1:
Hi Ellen @ekraffmiller , I've implemented the "unsaved changes" pop-up modal. It's ready for review.

I wanted to call out one behavior that might be confusing: The "Leave without saving" button prevents the changes from being submitted to the backend. However, the UI itself still temporarily saves the changes locally within the current tab. If the user navigates away and then back, the unsaved changes will still appear in the UI. I am not sure if we want this local persistence.

ekraffmiller:
> Hi Ellen @ekraffmiller , I've implemented the "unsaved changes" pop-up modal. It's ready for review.
> 
> I wanted to call out one behavior that might be confusing: The "Leave without saving" button prevents the changes from being submitted to the backend. However, the UI itself still temporarily saves the changes locally within the current tab. If the user navigates away and then back, the unsaved changes will still appear in the UI. I am not sure if we want this local persistence.

It looks good! I see what you mean, the changes are still there until you hit Save.  I think that's ok, we can change the warning to "If you leave now, your changes on this tab won't be saved".   The main thing is to make sure they save everything that they intend to save.

ekraffmiller (approved):
looks good!

jp-tosca:
There is currently a bug when you have custom terms of use, and the user clicks to see them, nothing happens, but if you refresh the page, it will load the terms of use.

Quick summary of the test:

1. Created a new dataset and published it.
2. Changed the terms of use to custom and saved the changes.
3. Did not publish this version of the dataset.

@ChengShi-1

PR #886, opened 2025-10-30
Update query parameters to Dataset&File Version Summaries use case 

### LINKED ISSUE #885: Update query parameters to Dataset&File Version Summaries use case
Based on JS-dataverse changes about Adding pagination query parameters to Dataset&File Version Summaries use case https://github.com/IQSS/dataverse-client-javascript/pull/395
This PR adds pagination support to the Dataset and File Version Summaries use cases by introducing optional limit and offset query parameters. This allows clients to retrieve version summaries in paginated chunks rather than loading all versions at once.

Front-end also needs to update these use cases accordingly

ChengShi-1:
Based on JS-dataverse changes about Adding pagination query parameters to Dataset&File Version Summaries use case https://github.com/IQSS/dataverse-client-javascript/pull/395
This PR adds pagination support to the Dataset and File Version Summaries use cases by introducing optional limit and offset query parameters. This allows clients to retrieve version summaries in paginated chunks rather than loading all versions at once.

Front-end also needs to update these use cases accordingly

- Closes #885 

it should be good if all checks pass, no ui change

no
yes, breaking change

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ This changelog follows the principles of [Keep a Changelog](https://keepachangel
 
 ### Changed
 
+- BREAKING CHANGE: add pagination support to the Dataset and File Version Summaries use cases by introducing optional limit and offset query parameters. #885
+
 - Use of the new `sourceLastUpdateTime` query parameter from update dataset and file metadata endpoints to support optimistic concurrency control during editing operations. See [Edit Dataset Metadata](https://guides.dataverse.org/en/6.8/api/native-api.html#edit-dataset-metadata) and [Updating File Metadata](https://guides.dataverse.org/en/6.8/api/native-api.html#updating-file-metadata) guides for more details.
 - Changed the way we were handling DATE type metadata field validation to better match the backend validation and give users better error messages. For example, for an input like “foo AD”, we now show “Production Date is not a valid date. The AD year must be numeric.“. For an input like “99999 AD”, we now show “Production Date is not a valid date. The AD year cant be higher than 9999.“. For an input like “[-9999?], we now show “Production Date is not a valid date. The year in brackets cannot be negative.“, etc.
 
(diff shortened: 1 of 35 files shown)


src/dataset/domain/repositories/DatasetRepository.ts
```diff
@@ -44,7 +44,11 @@ export interface DatasetRepository {
     paginationInfo: DatasetPaginationInfo
   ) => Promise<DatasetsWithCount>
   publish(persistentId: string, versionUpdateType: VersionUpdateType): Promise<void>
-  getDatasetVersionsSummaries: (datasetId: number | string) => Promise<DatasetVersionSummaryInfo[]>
+  getDatasetVersionsSummaries: (
```
ekraffmiller, line 48: This works but I think it would be good to pass a PaginationInfo object here, rather than the limit and offset, just to be consistent in how we are defining use cases that need pagination.  (You can define a new class, DatasetVersionPaginationInfo for that). 
 In looking at the code I realized I should update that too in getMyDataCollectionItems use case. 😊

CHANGELOG.md
```diff
@@ -15,6 +15,8 @@ This changelog follows the principles of [Keep a Changelog](https://keepachangel
 
 ### Changed
 
+- BREAKING CHANGE: add pagination support to the Dataset and File Version Summaries use cases by introducing optional limit and offset query parameters. #885
```
ekraffmiller: I don't think this counts as a breaking change for the frontend, because we don't have external users or developers depending on the use case (like we do in js-dataverse or the Dataverse API). So it's ok to describe just as a normal change/update.

ekraffmiller (commented):
looks good! just a suggestion about using a pagination param in the definition of the use case.

src/dataset/domain/repositories/DatasetRepository.ts
ChengShi-1: Good catch! I'll update it. I could also add paginationInfo to getMyDataCollectionItems here if you don't mind

ekraffmiller: thanks! Yes, that would be good if you want to add it to this PR...if you feel like it's too many extra changes, we can always add it as a separate issue.  
edit - Ok, I see you did it already, thanks!

ekraffmiller (approved):
looks good, approved!

ekraffmiller:
all tests pass and no merge conflicts, ok to merge

PR #901, opened 2025-12-15
881 notifications with paging

### LINKED ISSUE #881: Add paging to Notifications table
Update the notifications section to use paging, with infinite scroll.  The notifications will be marked as read (after a delay) when they are loaded into the table.

ekraffmiller:
Adds paging to Notifications UI

- Closes #881 

I created a separate "base branch" because 775 actually has all the commits for 775 and 881.


public/locales/en/account.json
```diff
@@ -53,6 +53,8 @@
     "userGuideLinkText": "User Guide",
     "demoServerLinkText": "Demo Site",
     "clearAll": "Clear All",
+    "clearAllOnThisPage": "Clear Notifications",
```
jp-tosca: I know this may make sense in the context it is used but do you think it would be good to have a key: "description" that is closer in meaning? "Clear notifications" is specific, while "clearAllOnThisPage" is very vague. 🤔

src/notifications/domain/models/NotificationsPaginationInfo.ts
```diff
@@ -0,0 +1,7 @@
+import { PaginationInfo } from '../../../shared/pagination/domain/models/PaginationInfo'
+
+export class NotificationsPaginationInfo extends PaginationInfo<NotificationsPaginationInfo> {
+  constructor(page = 1, pageSize = 10, totalItems = 0, itemName = 'result') {
```
jp-tosca, line 4: I think it would be better to have these parameters configurable somewhere outside the functions 😅

src/sections/account/notifications-section/NotificationsSection.tsx
```diff
@@ -51,47 +63,69 @@ export const NotificationsSection = ({ notificationRepository }: NotificationsSe
...
-                  <span className={styles['timestamp']}>{notification.sentTimestamp}</span>
+      <Stack gap={3} style={{ width: '100%' }}>
+        <Stack
+          direction="horizontal"
+          gap={2}
+          style={{ width: '100%', justifyContent: 'space-between', alignItems: 'center' }}>
+          <div>{t('notifications.displayingNotifications', { start, end, total })}</div>
+
+          {notifications.length > 0 && (
+            <Button
```
jp-tosca, line 92: It feels like the button’s behavior could be confusing. When there are multiple pages, the button says “Clear notifications”. The first time I saw that, I wasn’t sure—does it clear all notifications or just the ones on this page? After testing, I realized it only clears the current page, but that wasn’t obvious at first.

If there’s only one page, the text changes to “Clear all”, and in that case, it does clear everything.
This difference in wording and behavior feels inconsistent. I’d suggest making it static—something like “Clear all notifications”—and have it clear everything across all pages. We can chat more about this in standup.

src/sections/account/notifications-section/NotificationsSection.tsx
```diff
@@ -1,32 +1,44 @@
...
       .filter((n) => !n.displayAsRead && !readIds.includes(n.id))
       .map((n) => n.id)
-
+    console.log('in useEffect, unreadIds:', unreadIds)
     if (unreadIds.length > 0) {
       const timer = setTimeout(() => {
         void (async () => {
           await markAsRead(unreadIds)
           setReadIds((prev) => [...prev, ...unreadIds])
+          console.log('calling refetch after marking as read')
```
jp-tosca: Console log.

src/sections/account/notifications-section/NotificationsSection.tsx
```diff
@@ -1,32 +1,44 @@
...
+  )
+
   const [readIds, setReadIds] = useState<number[]>([])
 
   useEffect(() => {
     const unreadIds = notifications
       .filter((n) => !n.displayAsRead && !readIds.includes(n.id))
       .map((n) => n.id)
-
+    console.log('in useEffect, unreadIds:', unreadIds)
```
jp-tosca: Console log.

jp-tosca (commented):
Made some comments, we also discussed that in the future we probably should create a mechanism for when a component makes an operation that may generate a notification it should trigger a refresh to check the most recent number of notifications, otherwise the user won't see the notifications until next update, not a big deal but probably would improve the user experience.

src/sections/account/notifications-section/NotificationsSection.tsx
ekraffmiller: I like the idea of "Clear All" that deletes all notifications across all pages, but we don't have an API to support that - the delete API endpoint only deletes on notification at a time.  Some production accounts have thousands of emails, so deleting them all individually would be not ideal performance-wise.  We could add a "delete all notifications" API endpoint, to solve that, but for the time being, I can change the "Clear" to "Clear Current Page of Notifications", or 
"Clear Notifications 1-10". something like that. Does that sound good?

src/notifications/domain/models/NotificationsPaginationInfo.ts
ekraffmiller: They pageSize is configurable in the UI,  is that what you mean?  Also those values are just defaults, so when you instantiate the object, you can replace the default values with something else.
Screenshot 2026-01-09 at 11 22 28 AM

jp-tosca: What I was thinking is more about where this configuration is made (like a config file). I would rather have a single place on the application where these configurations exist, so if someone needs to change or configure, they don't have to go to the component itself and change it.

src/sections/account/notifications-section/NotificationsSection.tsx
jp-tosca: I think the 1-10 looks weird, and not sure if that is clear. Do you think we should leave it like that and just put a request later to change the endpoint to clear all, or create an endpoint that clears all?

jp-tosca: We talked about this and will add in the future an issue to improve this so we can delete all notifications, a new endpoint is required.

src/notifications/domain/models/NotificationsPaginationInfo.ts
jp-tosca: Since there are not many components that require configuration, we will leave it like this. If users make a request so that this can be configured, we will add some way to easily configure these.

jp-tosca (approved):

src/sections/account/notifications-section/NotificationsSection.tsx
```diff
@@ -1,27 +1,37 @@
+// TypeScript
```
ChengShi-1: Could you remove the `// TypeScript` here? and move the `import styles` in the end of all imports ?

ekraffmiller: thanks, good catch :)

ChengShi-1:
Hi @ekraffmiller , Thanks for all the work on the notification PR. It looks great!

I tested it locally with hundreds of notifications. I notice that when you clear notifications in the last page, it shows `no notification available`, but there are notifications in previous pages which I assume to be shown after a refresh. I attached the screen recording here to better demonstrate
https://github.com/user-attachments/assets/9894d5bc-7126-44bc-bb5e-8f5732267344

Also, some suggestions for UI
- About the time format, could you make it more readable like JSF?
image

- We could add a little gap between the text and notification icon here to make it look better, like `Dataverse Admin [3]` and `Notifications [3]` , which is similar to JSF as well
image
image

- also, we may need to add size labels to this PR and the issue.

ekraffmiller:
thanks for the QA @ChengShi-1! I will assign it to myself to make the fixes

ekraffmiller:
Hi @ChengShi-1 , I made the fixes to the paging and date format, but I'm not seeing the spacing issue in the notifications count...here is what it looks like for me:

Screenshot 2026-01-14 at 5 16 49 PM


Can you try rebuilding the design system? I think you might be missing updates in the Badge component.

ChengShi-1:
@ekraffmiller Thanks for the fixes! I see the badge after re-building design system, thanks. 

- Could you add a loading skeleton and an error alert to it? I see `return <div>Loading...</div>` and `return <div>{error}</div>` in NotificationsSection.tsx, it will be good if you could add a loading skeleton and an alert error info there
- Could you also update Changelog?

ekraffmiller:
@ChengShi-1 changes made, thank you

PR #921, opened 2026-02-09
Storage Quota/Upload Limit Integration

### LINKED ISSUE #897: Feature Request: Show storage/file quotas on the Upload Files page; disable Files to Add button when limits reached
The goal is to have the SPA Upload Files page mirror the behavior of the JSF equivalent as follows: 
When storage quota is present: 

Image

When the limit on the number of files in the dataset is present: 

Image

The upload button ('Select Files to Add' in the screenshots) gets disabled when either limit is maxed out. 
 
This pr in the main repo - IQSS/dataverse#11997 - finalizes the convenience API that can be used to obtain the information about any _remaining_ quota allocations for the dataset, when either one, or both of the limits are present (this will be in 6.9): 

```
/api/datasets/{id}/uploadlimits 
{
  "status": "OK",
  "data": {
     "uploadLimits": {
       "numberOfFilesRemaining": 20,
       "storageQuotaRemaining": 1048576
    } 
  }
}
```
When neither limit is present, the output will be like this: 
```
{
  "status": "OK",
  "data": {
     "uploadLimits": {} 
  }
}
```

The API requires authentication and the Edit permission on the dataset. 
Depends on js-dataverse use case: https://github.com/IQSS/dataverse-client-javascript/issues/409

ChengShi-1:
Show storage/file quotas on the Upload Files page

- Closes #897 

- I expect the upload limits will have more fields, but for now, it shows Storage Quota only

- When the file exceeds the storage quota, it would raise an error. It has been implemented before.
- The test coverage is not passing, but I solve it in other PR(Manage template)

- You may need to enable the setting and set upload limit through API first, to let the storage quotas set.
- Use the /settings API to enable or disable the enforcement of storage quotas that are defined across the instance via the following setting:
`curl -X PUT -d 'true' http://localhost:8080/api/admin/settings/:UseStorageQuotas`
- To set or change the storage allocation quota for a collection:
`curl -X PUT -H "X-Dataverse-key:$API_TOKEN" -d $SIZE_IN_BYTES $SERVER_URL/api/dataverses/$ID/storage/quota`
- Create a dataset, upload a file, then check the text above the upload box, "Select files or drag and drop into the upload widget. Maximum of 1,000 files per upload." should be shown and "Storage Quota: xx(the number you set minus the current file size) remaining" should also be shown if you set the number through PUT api above.

image

Yes, a screenshot above

Yes, changelog should be updated


src/dataset/domain/useCases/getDatasetUploadLimits.ts
ekraffmiller, line 1: This should use the DatasetRepository, rather than directly calling the jsDataverse function

src/sections/shared/file-uploader/file-upload-input/FileUploadInput.tsx
ekraffmiller, line 1: It would be good to have a FileUploadInput story that shows the file upload limit messages.

ekraffmiller (commented):
Looks good & works really nicely! just a couple of comments.

ekraffmiller (approved):
thanks for the updates, approved 👍

jp-tosca:
@ChengShi-1 @ekraffmiller 

I am looking into this. It seems that it is not very clear why this upload is failing: 

image

Is there a way to put a warning or somehow emphasize the text "Storage quota: 128 B remaining" and also provide a more informative message on why this failed? The text "There was an error uploading this file." doesn't provide enough feedback to people to understand why it failed, and the user relies on having to read the text above. 

Let me know what you think and if this would be an easy change.

ChengShi-1:
@jp-tosca Thanks for the QA and your feedback. I agree your suggestion makes sense here. 
However, the error message was implemented prior to this PR, so it isn’t covered in this change and I only addressed the text above. Also, since we may want to include this ASAP for the rollout. Can I just create a separate issue to handle this as a future optimization?

ekraffmiller:
@jp-tosca I think it's a good idea to show the error message from the API, but it's not a simple one line fix, since it has to go through some layers in the UI, so I think given our time frame is best to do that as a separate follow up issue.

jp-tosca:
The last item we discussed regarding to this PR and will be added as an improvement is the missing button "done".

PR #927, opened 2026-03-02
904 notifications refresh

### LINKED ISSUE #904: Feature Request/Idea:Notification refresh on operations that may generate a notification.

Right now, Dataverse generates notifications on certain operations like publishing a dataset. As it is right now, when the user publishes it, they won't see a notification until the next update (time-based) from the notifications UI.

One proposed solution would be to create a way to request a refresh and calling this after an operation that may generate a notification.

SPA User.

ekraffmiller:
refreshes the unread notifications list when these actions happen in the UI:
- View unread notifications 
- Create a dataset
- Publish a dataset 
- Create a Collection

- Closes #904 

There is an issue with the create dataset API, it isn't generating a notification: https://github.com/IQSS/dataverse/issues/12189

1. Perform the actions in the SPA that will create new unread notifications.
2. Test that the unread notifications icon in the header is correct. 
3. Go to the notifications tab, and view the unread notifications. 
4. Test that the unread notifications icon in the header is cleared when the notifications are marked as read.


tests/component/sections/account/NotificationsSection.spec.tsx
```diff
@@ -123,4 +134,13 @@ describe('NotificationsSection', () => {
 
     cy.get('@markAsRead').should('have.been.calledOnceWith', Cypress.sinon.match.number)
   })
+
+  it.only('does not render notifications display count when there are no notifications', () => {
```
ChengShi-1: Could you remove the "only" here?

ChengShi-1 (commented):
Looks good! The notification works well (except for the create dataset you mentioned). The test coverage is failing now, but I guess it should pass after removing "only" in one test case.

ChengShi-1 (approved):
Thanks for the change, approve

PR #940, opened 2026-03-26
Avoid prop-drilling for collection repository

### LINKED ISSUE #939: Avoid Prop-drilling for CollectionRepository
In order to avoid prop-drilling, we could use context to share repository instances.

We could create a custom hook,useRepositories( ) for passing use cases, wrap everything inside its provider component. Storybook and Test files should also be modified to match the same pattern of use context.

There are some sample codes from German of making these changes in this branch -> https://github.com/IQSS/dataverse-frontend/compare/develop...avoid-prop-drilling

ChengShi-1:
- the goal of the branch is to centralize repository injection at the React boundary and remove UI prop-drilling, not to eliminate explicit repository dependencies from domain/use-case code
- moduleResolution: "bundler" is intentional and aligns TypeScript with the Vite build; if it surfaces more deep imports, those imports should be treated as bugs rather than reverted
- the branch is mechanically broad because the provider change touches factories, pages, stories, and tests together

- Closes #939 

- "moduleResolution": "node" is deprecated  `Error: tsconfig.json(14,25): error TS5107: Option 'moduleResolution=node10' is deprecated and will stop functioning in TypeScript 7.0. Specify compilerOption '"ignoreDeprecations": "6.0"' to silence this error. `
Official doc of ts: https://www.typescriptlang.org/tsconfig/#moduleResolution  

Why bundler exposed that AlertVariant import:
- bundler respects package public exports. @iqss/dataverse-design-system only exports its root API.
- Deep-importing dist/components/... was relying on package internals, so bundler rejected it.

There is no ui change or functional change could be tested

No

Yes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ This changelog follows the principles of [Keep a Changelog](https://keepachangel
 - Renamed dataset template fetch/create use cases and DTOs to `getTemplatesByCollectionId` and `CreateTemplateDTO` for API alignment, and added new `getTemplate` and `deleteTemplate` use cases for retrieving a single template by ID and deleting templates.
 - Added disclaimer text and custom popup text to Publish Dataset modal for better communication of the implications of publishing a dataset. The text can be configured through the runtime configuration.
 - Dataset page Terms tab title now depends on permissions: users with dataset update permission see `Terms and Guestbook`, and read-only users see `Terms`.
+- Avoided prop-drilling for collection repository, so used context to share epository instances.
 
 ### Fixed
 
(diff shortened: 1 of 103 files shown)


rtreacy:
Removes prop drilling, standardizes repository access, and simplifies tests/stories. Only issue is reduces test coverage slightly to 94.92% which brings it below global threshold

rtreacy (approved):
Uses RepositoryProvider to manage CollectionRepsoitory. Reduces prop-drilling across many components, simplifies component signatures, Improves maintainability. Test coverage over 97%

PR #967, opened 2026-04-24
use config.js from code base when deploying

ekraffmiller:
Update the build scripts so they use the bannerMessage defined in config.js, rather than creating a separate config object.

This is needed so we can deploy the correct banner message in the github deploy actions.


ChengShi-1 (approved):
looks good

PR #984, opened 2026-05-04
use new modern url in banner message

ekraffmiller:
The banner message needed to be updated to use the new `modern` url.  Even though this is configurable at runtime, it's easier to create another war file than to update it directly in the Payara environment.


ChengShi-1 (approved):
looks good

PR #1018, opened 2026-07-15
Update Payara version in deploy actions

ekraffmiller:
fix payara path, should be payara7


jp-tosca (approved):
Looks good, jus the change of version

PR #1027, opened 2026-07-22
Search results map

### LINKED ISSUE #580: Feature Request/Idea: Display dataset search results on a geographical map
Add the option to display search results (datasets) that have location information (coordinates in the metadata) on an interactive map. The map could either be an alternative 'search results view', instead of the paged list view, or it could be on the side along with the facets. 

This feature request comes from the archaeological community that uses our Dataverse archival system specfic for archaeological research data (excavation results etc. ). 
Most dataset in that archive ([https://archaeology.datastations.nl](https://archaeology.datastations.nl)) have a location specified in the metadata which was used in our previous online system (non-Dataverse) to display the dataset locations on a map. 
 
As a prove of concept (PoC) I developed javscript code that gets inserted into the custom footer (html) file. This script will add 'tabs' for the 'listview' and the 'mapview' just above the result listing area on the dataverse page. When the 'mapview' tab is selected the map (Leaflet with OSM tiles) is displayd with markers for every dataset that has coordinates. 

Explanantion and code of the PoC can be found here:
https://github.com/Dans-labs/landscape-analysis-GeoMapDataverseIntegration/blob/main/experiments/MapViewExtention.md

Some screenshots from our test site:

test-archaeology-listview

test-archaeology-mapview

Demo screencast:

https://github.com/user-attachments/assets/3483bb16-cff8-4d78-973e-51b0da7000f0

One of the main challenges is that the community will have those coordinates in different places (custom metadata) and also in different systems, so configuring that is not straightforward in TSV files. 

Configuration would be to specify; 
- The subverse/collection for which you want it (default root)
- The query addition needed to retrieve the metadata fields (from a custom block). 
- The script code (or some other 'pluggable' code) to extract and convert that metadata into geojson

Another changeling thing is the relative long time it takes to get the results from the search API and the fact that the total number of results might be high; 150k hits on our archive for instance would require 150 http requests. Also, 150k markers on the map is probably making it unusable.

julian-schneider:
Adds a browsable map of all search results with geospatial information. It is located in a tab on the search results page, as an alternative to the normal list of search results. The map displays one pin per dataset, which can be clicked for a small preview that leads the user to the details page. Clusters of pins are grouped to declutter the area.

This implementation is entirely based on the proof of concept by @PaulBoon described in detail [here](https://github.com/Dans-labs/landscape-analysis-GeoMapDataverseIntegration/blob/main/experiments/MapViewExtention.md). That PoC uses a custom footer to inject the map into the classic UI, while this implementation is based on `react-leaflet`.

This PR is a work-in-progress draft version. Some open issues I'm aware of:
- [x] The geospatial metadata block contains any number of bounding boxes, while the map is designed to display one point. -> The map now displays a marker at the averaged center of the dataset's bounding boxes. When the marker is clicked, the bounding boxes will be displayed along with the dataset info popup.
- [x] The initial prototype didn't pass the user's auth token when fetching geospatial datasets -> now does, but in a hacky manner
- [ ] Performance: Map population time mostly depends on the search API call to fetch the datasets with geolocations. In simple tests, this speed varied strongly between installations (0.2s - 3.0s for 50 datasets). `query_entities=false` would create a substantial speedup, but is currently not applicable because it removes the geolocations from the returned metadata.
- [x] i18n should be used -> done
- [x] Pagination loads the first N search results to display on the map, irrespective of whether these have geolocations, so the first batch size varies and may even be empty -> now filters for datasets with geolocations

- Closes #580 

Populate an installation with datasets filling out the official geospatial metadata block, and the map should display them.

Yes. The map was entirely based on the design of the previously-mentioned [PoC](https://github.com/Dans-labs/landscape-analysis-GeoMapDataverseIntegration/blob/main/experiments/MapViewExtention.md).

Yes


julian-schneider:
Screenshot:
image

PaulBoon:
@julian-schneider Great work! 
If you want to look at the performance you could have look at the stress tests that I did some time ago; you could probably reuse the test data generating Python code and possibly the Jmeter tests. 
https://github.com/DANS-KNAW/dataverse-dataset-geomap-view/tree/main/examples/testdata/stresstests

qqmyers:
FWIW: Using ?query_entities=false in the Dataverse API is a factor or 10x faster, so unless you needed the added info, I'd definitely add it.

julian-schneider:
Made some updates, here's a short video showcasing the behavior :world_map: 

[map-tab-status.webm](https://github.com/user-attachments/assets/8dc817f9-c8ad-4f51-a6cd-1d3ab5679a79)

Note that each dataset's marker is placed in the center of its bounding box (averaged, if it has more than one). These bounding boxes come from the geospatial metadata block.

Initially, the map loads the 50 geospatial datasets that are the most relevant results of the search query. "Load more" fetches the next 50 datasets and adds them into the view.

julian-schneider:
Regarding performance:

Thanks for linking your stuff @PaulBoon! I've used the data generation tool to produce an initial set of around 200 datasets for testing purposes.

So far I've only executed the project locally using `dev-env/run-env.sh`. The API is taking a long time to answer search queries, taking about 10s to fetch the current page size of 50 datasets.

@qqmyers I'm not sure why, but `query_entities=false` currently does not seem to have an effect when running the project locally, as mentioned above. These 2 requests:
- `api/search?q=*&type=dataset&subtree=root&start=0&per_page=50&fq=geolocation:*&query_entities=false`
- `api/search?q=*&type=dataset&subtree=root&start=0&per_page=50&fq=geolocation:*&query_entities=true`

yield the same API response, and take about the same time. Also, to populate the map I definitely need the bounding boxes, so I need to use `metadata_fields=geospatial:*`. And I assume that probably prevents me from using `query_entities=false` anyway, right?

qqmyers:
query_entities was restored as a working parameter in Feb. 2025 so Dataverse releases before then won't use it. If that's not the issue, I can't think of anything off-hand. As for getting the bounding box, those fields are in solr - I'd guess that means you don't need to query the entities, but I'm not sure w/o checking. I guess it would be easy to check once the issue with not being able to see a difference with query_entities true/false is resolved.

julian-schneider:
(Ah yes when executing `run-dev.sh unstable`, I accidentally ran quite an old image (version 6.5) where `query_entities` didn't function, thanks for pointing me in the right direction @qqmyers :+1:)

With that fixed, now I'm getting the drastic speed-up from `query_entities=false`. Sadly, using this does cause the geospatial information to be skipped, even with `metadata_fields` usage:

The query `api/search?q=*&type=dataset&subtree=root&start=0&per_page=50 &metadata_fields=geospatial:*&fq=geolocation:*&query_entities=false` responds without `metadataBlocks`, e.g. this item:

```json
{
  "name": "Test dataset 27  with 2 boxes 3ee7538a-8b72-461d-8221-7dc29a05f2bd",
  "type": "dataset",
  "url": "https://doi.org/10.5072/FK2/VOS4AT",
  "global_id": "doi:10.5072/FK2/VOS4AT",
  "description": "test",
  "published_at": "2026-08-14T10:47:38Z",
  "publisher": "Root",
  "citationHtml": "Admin, Dataverse, 2026, \"Test dataset 27 with 2 boxes 3ee7538a-8b72-461d-8221-7dc29a05f2bd\", <a href=\"https://doi.org/10.5072/FK2/VOS4AT\" target=\"_blank\">https://doi.org/10.5072/FK2/VOS4AT</a>, Root, V1",
  "identifier_of_dataverse": "root",
  "name_of_dataverse": "Root",
  "citation": "Admin, Dataverse, 2026, \"Test dataset 27 with 2 boxes 3ee7538a-8b72-461d-8221-7dc29a05f2bd\", https://doi.org/10.5072/FK2/VOS4AT, Root, V1",
  "publicationStatuses": [
    "Published"
  ],
  "authors": [
    "Admin, Dataverse"
  ]
},
```

As far as I can tell, [SolrSearchResult.java](https://github.com/IQSS/dataverse/blob/develop/src/main/java/edu/harvard/iq/dataverse/search/SolrSearchResult.java) and [SolrSearchServiceBean.java](https://github.com/IQSS/dataverse/blob/develop/src/main/java/edu/harvard/iq/dataverse/search/SolrSearchServiceBean.java) decide which fields from Solr are added into the API response. Do you think we could get the geospatial information in there? Or would it be possible to add a parameter that allows specifying which fields from the Solr schema should get returned (so, similar to `metadata_fields`)?

qqmyers:
Bummer. In general, I can’t think of any reason we couldn’t return more fields from solr, someone would just have to do the work.

The only thing I can think to make sure you’re search is as fast as currently possible is to use the [geo_point/geo_radius parameters](https://guides.dataverse.org/en/latest/api/search.html#parameters) to just find the datasets that would be in view.

PaulBoon:
@julian-schneider About the slowness, we run the map with 100 results per request (for more) and it works fine. It is not fast, but does not take 10 seconds, unless we have lost of simultaious users. Just have a look at https://archaeology.datastations.nl/. 
So maybe something else is slowing things down, have you inspected the page with the dev tools to see where it spending time loading?

julian-schneider:
@PaulBoon The majority of waiting time is just the `/search` API call itself. And these super slow 10s calls only occurred when using the dataverse container ran by using `dev-env/run-env.sh` in this repo.

I've roughly compared the search API speed of a few different live Dataverse instances. When requesting `api/search?q=*&per_page=50&query_entities=true`:

- On the cluster I'm deploying Dataverse to, it takes ~3.0s.
- [dataverse.harvard.edu](https://dataverse.harvard.edu/api/search?q=*&per_page=50&query_entities=true) takes ~1.8s
- [data.fz-juelich.de](https://data.fz-juelich.de/api/search?q=*&per_page=50&query_entities=true) takes ~1.0s
- [archaeology.datastations.nl](https://archaeology.datastations.nl/api/search?q=*&per_page=50&query_entities=true) takes ~0.2s (impressive!)

Considering this major difference, I'm not surprised that you're happily fetching pages of size 100 :smile: Did you do any performance optimizations on your instance that you would recommend?

Otherwise I think the best I can do short-term is to keep the page size down, which makes a large difference with `query_entities=true`, and for long-term, look into whether the bboxes could be returned even with `query_entities=false`.

julian-schneider:
@pdurbin Since I wish to also reuse this map in the custom frontend of my installation (the one I described [here](https://dataverse.zulipchat.com/#narrow/channel/376593-geospatial/topic/search.20results.20on.20a.20map/near/609868503)), I have separated some parts of its code and created a storybook entry for the display component. However, the project currently only bundles and publishes the `packages/design-system/` components. Do you think other components, in my case `CollectionMapUI`, could also be published in the future?

pdurbin:
@julian-schneider we have a weekly frontend meeting. I'll ask!

ChengShi-1:
> @pdurbin Since I wish to also reuse this map in the custom frontend of my installation (the one I described [here](https://dataverse.zulipchat.com/#narrow/channel/376593-geospatial/topic/search.20results.20on.20a.20map/near/609868503)), I have separated some parts of its code and created a storybook entry for the display component. However, the project currently only bundles and publishes the `packages/design-system/` components. Do you think other components, in my case `CollectionMapUI`, could also be published in the future?

Hi @julian-schneider, thanks for splitting this out and adding a Storybook entry, it’s very helpful. 

For reusable UI components, we usually publish them as part of the design system, so people can explore and reuse them there. Is that what you mean, or are you suggesting a totally separate package alongside `packages/design-system`? We generally don’t create separate packages for UI components since the `packages/design-system` should be easy to reuse, and a separate package would add maintenance overhead

julian-schneider:
Hey @ChengShi-1, thanks for your input about this. If you say that the design-system is the appropriate place for this map, that's great to hear and there's no reason to create any other package.

I have now moved the map display component into design-system for bundling :+1:

PR #1043, opened 2026-08-07
skip Chromatic checks for forks

pdurbin:
## What this PR does / why we need it:

To avoid confusion for people make PRs who don't have write access to the upstream IQSS repo (who have forked).

## Which issue(s) this PR closes:

- Closes #1042

## Special notes for your reviewer:

There's a surprising variety of solutions out there. In the end I went with a slightly simplified version of the one we use here: https://github.com/IQSS/dataverse/blob/v6.11/.github/workflows/copy_labels.yml#L10

Codex and Claude both came up with this:

`if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository`

Googling yields a variety of methods as well.

## Suggestions on how to test this:

- Merge this PR.
- Make a PR from IQSS and check to make sure the Chromatic checks DO run. (No regression.)
- Make a PR from a fork and check to make sure the Chromatic checks DON'T run. (A good fix.)

## Does this PR introduce a user interface change? If mockups are available, please link/include them here:

No.

## Is there a release notes or changelog update needed for this change?:

No.

## Additional documentation:

None.

PR #1048, opened 2026-08-17
Fix human-readable publication relation type labels

### LINKED ISSUE #1041: Relation Type menu options text in the Related Publication not displaying properly; no spaces between words
All the Relation Type options in the Related Publication section of dataset metadata is not displaying properly in the Modern UI. There is no spaces in the text. This error is also displayed out side of the edit metadata window. The text is properly spaced in the JSA DV version 

- When does this issue occur?
When editing and viewing metadata on a dataset where the Related Publication field uses an option from the Relation Type dropdown menu.

I noticed it on this dataset page: https://dataverse.harvard.edu/modern/datasets?persistentId=doi:10.7910/DVN/IKIVID

- What happens?

- To whom does it occur (all users, curators, superusers)?
Superuser and not logged in at all. Didn't test other login types.

I expect the phrases in the Relation Type dropdown menu have a space between each word.

v. 6.10.1 build iqss-4|frontend version: 371cf9773

**Screenshots:**

<kbd>Image</kbd>

<kbd>Image</kbd>

How the same text is displayed in the JSA version:
<kbd>Image</kbd>

No matter the issue, screenshots are always welcome.

To add a screenshot, please use one of the following formats and/or methods described here:

- https://help.github.com/en/articles/file-attachments-on-issues-and-pull-requests

youseihuayu-wonderful:
## Problem

The modern UI uses raw controlled-vocabulary values as display labels for
`publicationRelationType`. This causes values such as `IsSupplementedBy` to
appear instead of the human-readable `Is Supplemented By`.

## Implementation

- Add an explicit, field-scoped label mapping for all six publication relation
  types.
- Use separate `{ value, label }` options in the dataset metadata edit form.
- Apply the same labels when formatting relation types in the dataset metadata
  view.
- Fall back to the original value when an unknown relation type is encountered.
- Leave all other controlled-vocabulary fields unchanged.

## Backward compatibility

This is a presentation-only change. Form submissions, stored metadata, API
values, and DataCite-compatible values continue to use the original raw values
such as `IsSupplementedBy`.

Existing datasets receive the corrected labels at render time without migration,
reindexing, or resaving.

## Validation

- Added coverage for all six relation type mappings.
- Verified unknown-value fallback.
- Verified unrelated controlled vocabularies remain unchanged.
- Verified the edit dropdown displays labels while submitting raw values.
- Verified dataset metadata rendering displays human-readable labels.
- Ran 80 relevant Cypress component and regression tests.
- TypeScript typecheck, ESLint, Prettier, and production build passed.

Closes #1041


tests/component/sections/shared/dataset-metadata-form/Vocabulary.spec.tsx
```diff
@@ -0,0 +1,78 @@
...
+      .select('IsSupplementedBy')
+      .should('have.value', 'IsSupplementedBy')
+    cy.findByRole('button', { name: 'Submit' }).click()
+
+    cy.get('@onSubmit').should('have.been.calledOnce')
+    cy.get('@onSubmit')
+      .its('firstCall.args.0.citation.publication.0.publicationRelationType')
+      .should('equal', 'IsSupplementedBy')
+  })
+})
```
pdurbin, line 120: I'm just leaving this comment here at the bottom, regarding these failing checks:

Image

The first two should be fixed by this PR:

- #1043

The next one doesn't have a PR but I just opened this issue:

- #1051 

For the last one, I'm seeing weird stuff like this:

`<<<CYPRESS.STDERR.END>>>[6475:0817/203901.677876:ERROR:dbus/bus.cc:408] Failed to connect to the bus: Could not parse server address: Unknown address type (examples of valid types are "tcp" and on UNIX "unix")`

ChengShi-1: Fork-based E2E runs fail because the workflow needs S3_SECRET_KEY, and GitHub does not provide secrets to pr workflows originating from forks. These e2e test files are expected to fail 

- FileUpload.spec.ts — explicitly tests direct S3 upload.
- FileJSDataverseRepository.spec.ts — creates files, then tests file operations and downloads.
- Dataset.spec.tsx — many file creation, upload, and download flows.
- File.spec.tsx — creates files and verifies file download behavior.
- EditFileMetadata.spec.tsx — requires a real stored file.
- CollectionItemsPanel.spec.ts — creates datasets with files.

I'd suggest the reviewer to run e2e test locally to see if they're passing

src/metadata-block-info/domain/models/ControlledVocabularyValueLabel.ts
```diff
@@ -0,0 +1,23 @@
...
+  IsCitedBy: 'Is Cited By',
+  Cites: 'Cites',
+  IsSupplementTo: 'Is Supplement To',
+  IsSupplementedBy: 'Is Supplemented By',
+  IsReferencedBy: 'Is Referenced By',
+  References: 'References'
+}
+
+/**
+ * Returns the human-readable label for controlled vocabulary values that need
```
ChengShi-1: We could remove the comments here

src/metadata-block-info/domain/models/ControlledVocabularyValueLabel.ts
```diff
@@ -0,0 +1,23 @@
...
+ * presentation-specific formatting. The original value remains the value used
+ * by forms, APIs, and stored metadata.
+ */
+export function getControlledVocabularyValueLabel(fieldName: string, value: string): string {
```
ChengShi-1: Could you move this function to /MeatadataBlockInfo.ts and remove this file? It's good to manage everything related to metadata info inside a single file.

ChengShi-1: Could you change the name `getControlledVocabularyValueLabel` to `getPublicationRelationLabel` or something similar, since this function is only use for `PUBLICATION_RELATION_TYPE_LABELS` instead of all controlled vocabularies?

src/sections/shared/form/DatasetMetadataForm/MetadataForm/MetadataBlockFormFields/MetadataFormField/Fields/Vocabulary.tsx
```diff
@@ -7,6 +7,7 @@ import { MetadataFieldsHelper } from '../../../../MetadataFieldsHelper'
 import { type CommonFieldProps } from '..'
 import { CustomInstructionsEditor } from '../CustomInstructionsEditor'
 import styles from '../index.module.scss'
+import { getControlledVocabularyValueLabel } from '@/metadata-block-info/domain/models/ControlledVocabularyValueLabel'
```
ChengShi-1: As a standard format, we usually put `import styles...` as the last line of the whole import section.

src/metadata-block-info/domain/models/ControlledVocabularyValueLabel.ts
```diff
@@ -0,0 +1,23 @@
+const PUBLICATION_RELATION_TYPE_FIELD_NAME = 'publicationRelationType'
+
+const PUBLICATION_RELATION_TYPE_LABELS: Readonly<Record<string, string>> = {
```
ChengShi-1: The labels are hard-coded English strings, so they won’t automatically localize if this field becomes translated later. Could you move these labels into the locale resources and resolve them with `t(...)` when rendering?

tests/component/metadata-block-info/domain/models/ControlledVocabularyValueLabel.spec.ts
```diff
@@ -0,0 +1,30 @@
+import { getControlledVocabularyValueLabel } from '@/metadata-block-info/domain/models/ControlledVocabularyValueLabel'
+
+describe('getControlledVocabularyValueLabel', () => {
+  const relationTypeLabels = {
```
ChengShi-1: This hard-coded `const` could be imported

ChengShi-1 (changes requested):
Hi Sy, thanks for your contribution! I left some comments here. The main concern is the hard-coded string, which could be made dynamic to support internationalization.

src/metadata-block-info/domain/models/ControlledVocabularyValueLabel.ts
youseihuayu-wonderful: Removed along with the dedicated helper file.

src/metadata-block-info/domain/models/ControlledVocabularyValueLabel.ts
youseihuayu-wonderful: Done — moved the helper to MetadataBlockInfo.ts, renamed it to getPublicationRelationLabel, and removed the old file.

src/sections/shared/form/DatasetMetadataForm/MetadataForm/MetadataBlockFormFields/MetadataFormField/Fields/Vocabulary.tsx
youseihuayu-wonderful: Done — the styles import is now the last import.

src/metadata-block-info/domain/models/ControlledVocabularyValueLabel.ts
youseihuayu-wonderful: Done — the labels now live in the English and Spanish shared locale resources, and both rendering paths resolve them through t(...).

tests/component/metadata-block-info/domain/models/ControlledVocabularyValueLabel.spec.ts
youseihuayu-wonderful: Done — the tests now import PUBLICATION_RELATION_TYPE_VALUES from MetadataBlockInfo.ts.

tests/component/sections/shared/dataset-metadata-form/Vocabulary.spec.tsx
youseihuayu-wonderful: Updated with the latest develop. Both Chromatic checks are now skipped as expected. The label-copy workflow issue remains tracked in #1051, and I did not change the fork E2E workflow in this PR.

youseihuayu-wonderful:
@ChengShi-1 Thanks for the review. I addressed all comments which passed on immediate isolated rerun (38/38). Could you please take another look?

src/metadata-block-info/domain/models/MetadataBlockInfo.ts
```diff
@@ -91,3 +93,21 @@ export type MetadataFieldInfo = Pick<
...
+export const PUBLICATION_RELATION_TYPE_VALUES = [
+  'IsCitedBy',
+  'Cites',
+  'IsSupplementTo',
+  'IsSupplementedBy',
+  'IsReferencedBy',
+  'References'
+] as const
+
+export function getPublicationRelationLabel(value: string, t: TFunction): string {
```
ChengShi-1: Could you change the `t: TFunction` in param to `t: (key: string) => string`?

ChengShi-1, line 107: Sorry I didn't mention it from last review. After another look, I think the way we get the label is not to create a helper like this. The usual pattern in this codebase is to translate the value directly in the component. 

We could remove this function by adding the follwing to `Vocabulary.tsx`, so we wouldn't need to pass the helper function through multiple files.

`label:
  name === PUBLICATION_RELATION_TYPE_FIELD_NAME &&
  PUBLICATION_RELATION_TYPE_VALUES.some((value) => value === option)
    ? t(`publicationRelationTypes.${option}`)
    : option`

`MetadataBlockInfo.spec.ts` and some related tests should be removed as well.
Before you push the commit, don't forget to run `npx prettier --write .` to automatically fix format error

ChengShi-1 (changes requested):

PR #1057, opened 2026-08-21
Use LocalStack for development and E2E S3 storage

### LINKED ISSUE #1053: Make GitHub Actions E2E tests independent of S3 secrets
## Summary
Fork-based pull requests cannot run the full E2E suite because the GitHub Actions workflow requires `S3_ACCESS_KEY` and `S3_SECRET_KEY`. 

S3-dependent tests include:
- `FileUpload.spec.ts`
- `FileJSDataverseRepository.spec.ts`
- `Dataset.spec.tsx`
- `File.spec.tsx`
- `EditFileMetadata.spec.tsx`
- `CollectionItemsPanel.spec.ts`

We could make E2E storage self-contained and independent of repository secrets—for example, use a local service instead.

### LINKED ISSUE #1056: As a contributor, I'd like the dev environment to allow uploading files (LocalStack instead of AWS with credentials)
During the frontend meeting this week we talked about how e2e tests fail for contributors because their forks don't have the credentials necessary to upload files. This is now captured in this issue:

- #1053 

On a related note, contributors who are setting a dev environment may notice that they can't upload files. This is because the dev environment requires credentials for a S3 bucket on AWS that IQSS pays for. The dev guide explains it somewhat obliquely this way:

```
# copy the .env.example file to .env
# To test file upload, update the .env file with S3 credentials
$ cp .env.example .env
```

I just checked with @ChengShi-1 and she's fine with switching to LocalStack for the dev environment. This what we do for the backend and as far as we know LocalStack is completely compatible with real AWS S3.

I included links in #1053 about to the PR where we added LocalStack to the backend but here it is are again:

- https://github.com/IQSS/dataverse/pull/10044

You'll find MinIO in that pull request as well, but we later removed it in this one:

- https://github.com/IQSS/dataverse/pull/12394

youseihuayu-wonderful:

Replaces the AWS S3 bucket used by the containerized development environments and E2E tests with a local [LocalStack](https://www.localstack.cloud/) container.

This makes file uploads self-contained for contributors and allows fork-based pull requests to run the complete E2E suite without access to IQSS AWS credentials.

The changes:

- Add an S3-only `dev_localstack` service to the standard and Shibboleth development environments.
- Configure Dataverse to use `default/default` development credentials, bucket `dataverse`, path-style URLs, and `http://s3.localhost:4566`.
- Wait for the LocalStack bucket to be healthy before starting Dataverse.
- Add a shared initialization script that creates the bucket and configures browser CORS, including exposed `ETag` headers for multipart uploads.
- Remove `S3_ACCESS_KEY` and `S3_SECRET_KEY` from the development environment examples and GitHub Actions workflow.
- Document that local file storage requires no AWS account or credentials.

The backend LocalStack setup uses the container-only hostname `localstack`. The frontend uses the `s3.localhost` network alias because direct-upload and download URLs must be reachable under the same hostname from both Dataverse containers and the developer's browser. The browser also requires the additional CORS configuration.

- Closes #1053
- Closes #1056

- LocalStack is pinned to `localstack/localstack:4.2.0`, matching the Dataverse backend development environment.
- Port 4566 is bound to `127.0.0.1` only.
- Storage is intentionally ephemeral on a 128 MB tmpfs. Recreating the environment discards LocalStack objects, as documented in `DEVELOPER_GUIDE.md`.
- `S3_ACCESS_KEY`, `S3_SECRET_KEY`, the paid `beta-dataverse-direct` bucket, and the real AWS endpoint are no longer referenced anywhere in the repository. The repository S3 secrets can be removed after this is merged.
- The Shibboleth-specific patch changes only storage configuration and startup ordering; it does not change its authentication flow. Its direct upload and download redirects were verified at the API level, but the browser Shibboleth SSO flow was not retested as part of this change.

The final PR head passed the complete fork-based pull request workflow:

- https://github.com/IQSS/dataverse-frontend/actions/runs/32533842337
- E2E: 146 tests, 139 passing, 7 pre-existing pending, 0 failing.
- Component: 267 passing, 0 failing.
- All six S3-dependent specs listed in #1053 passed, including all 38 `Dataset.spec.tsx` tests.

The same storage setup also passed an earlier push workflow on the fork, which has no repository secrets configured:
https://github.com/youseihuayu-wonderful/dataverse-frontend/actions/runs/32513442649

Local verification was also run against both Compose environments:

1. Copy the relevant `.env.example` to `.env`.
2. Start `dev_localstack` with Docker Compose.
3. Confirm the container becomes healthy and bucket `dataverse` exists.
4. Confirm its CORS policy permits `PUT` and exposes `ETag`.
5. Confirm `s3.localhost` resolves inside the Compose network.
6. Generate a presigned URL inside that network and upload to it from the host.

Both environments passed all six checks. Compose rendering, Bash syntax, Prettier, TypeScript, ESLint, Stylelint, and the production build were also checked locally.

No.

No. This changes development and CI infrastructure only.

## Additional documentation:

- `DEVELOPER_GUIDE.md` documents the LocalStack endpoint, hostname fallback, and ephemeral storage behavior.
- `dev-env/shib-dev-env/README.md` documents that its file uploads no longer require AWS credentials.


pdurbin (approved):
This works on my machine! I commented out the S3 keys and I was still able to upload a file!

Image

I'm approving this but I'll keep it in "ready for triage" so we can talk more about it.

This opens up the repo for contribution in a huge way! Thanks, @youseihuayu-wonderful!

dev-env/docker-compose-dev.yml
```diff
@@ -92,6 +97,32 @@ services:
...
+      - '127.0.0.1:4566:4566'
+    environment:
+      SERVICES: s3
+      S3_BUCKET_NAME: dataverse
+    healthcheck:
+      test: ['CMD-SHELL', 'awslocal s3api head-bucket --bucket "$$S3_BUCKET_NAME"']
+      interval: 5s
+      timeout: 5s
+      retries: 20
+      start_period: 10s
```
ChengShi-1: The healthcheck only confirms the bucket exists, not that put-bucket-cors succeeded, so a CORS failure would leave the container reporting healthy while browser uploads silently fail preflight. 

The retry budget (~110s total) may also be tight for a cold LocalStack start. Suggest checking a completion marker written by init-s3.sh after CORS setup, and giving it a bit more headroom:

      test: ['CMD-SHELL', 'test -f /tmp/.s3-init-complete && awslocal s3api head-bucket --bucket "$$S3_BUCKET_NAME"']
      interval: 5s
      timeout: 5s
      retries: 30
      start_period: 15s

Needs a matching one-line addition to `dev-env/localstack/init-s3.sh` (after the `put-bucket-cors` call): `touch /tmp/.s3-init-complete` 
Same block also exists in `dev-env/shib-dev-env/docker-compose-dev.yml` and should get the same fix.

dev-env/localstack/init-s3.sh
```diff
@@ -0,0 +1,22 @@
...
+  --cors-configuration '{
+    "CORSRules": [
+      {
+        "AllowedHeaders": ["*"],
+        "AllowedMethods": ["GET", "HEAD", "PUT", "POST", "DELETE"],
+        "AllowedOrigins": ["*"],
+        "ExposeHeaders": ["ETag"]
+      }
+    ]
+  }'
```
ChengShi-1, line 24: `touch /tmp/.s3-init-complete`

ChengShi-1 (commented):
Looks great, thank you, Sy! just a few comments here

dev-env/docker-compose-dev.yml
```diff
@@ -92,6 +97,32 @@ services:
...
+    networks:
+      dataverse:
+        # Use the same hostname from browsers and containers so presigned URLs remain valid.
+        aliases:
+          - s3.localhost
+    volumes:
+      - ./localstack:/etc/localstack/init/ready.d
+    tmpfs:
+      - /var/lib/localstack:mode=770,size=128M,uid=1000,gid=1000
```
pdurbin, line 128: At Triage Tuesday today we talked briefly about how all the other data (for Postgres and Solr for example) is stored in a directory called "docker-dev-volumes".

How much work would it be to include the LocalStack data in that directory?

By the way, the "docker-dev-volumes" is commonly deleted. See the dev-env/rm-env.sh script. However, some of us, myself included, sometimes comment out the `rm -rf ./docker-dev-volumes` line so we can stop the environment and not lose all the data.

If it's too much work, after we merge this, I'd be happy to create an issue for the future.

youseihuayu-wonderful: Thanks, Philip. I looked into this. Moving the mount under `docker-dev-volumes` would be easy, but the persistence itself does not work with our current Community image just by mounting `/var/lib/localstack`.

I tested `localstack/localstack:4.2.0` with `PERSISTENCE=1`, an `ON_SHUTDOWN` snapshot strategy, and a bind-mounted state directory. After creating an S3 object, gracefully stopping/removing the container, and recreating it with the same directory, the bucket initializer ran again but the object was gone. The directory contained LocalStack cache files but no state snapshot. The official persistence setup uses the Pro image and an auth token: https://docs.localstack.cloud/aws/capabilities/state-management/persistence/

Since adding an auth token would reintroduce a secret requirement for forks and contributors, I don't think we should add a mount that looks persistent but does not actually preserve S3 data. A Community-compatible backup/restore mechanism would be a larger follow-up. I recommend keeping the documented ephemeral setup in this PR and opening the follow-up issue you suggested. Does that sound good?

youseihuayu-wonderful: @pdurbin  Hi Philip. I tested this with the Community image used by this PR. Mounting LocalStack’s
 data directory under `docker-dev-volumes` is easy, but the mount alone does not preserve S3 objects
 when the container is recreated.

   LocalStack’s official persistence setup requires the Pro image and an auth token, so a
 credential-free Community solution would require additional backup/restore implementation and
 testing. I think it makes sense to keep the current temporary behavior in this PR and address
 persistence in a follow-up issue after merge.

dev-env/docker-compose-dev.yml
youseihuayu-wonderful: @ChengShi-1 
 Implemented in both Compose files as suggested. Could you please rerun the failed E2E job? I
 don’t have permission to rerun it.
 https://github.com/IQSS/dataverse-frontend/actions/runs/32882924859/job/97916490555

ChengShi-1: Thanks @youseihuayu-wonderful , I just rerun it

dev-env/docker-compose-dev.yml
pdurbin: Yeah at https://www.localstack.cloud/pricing they call it "local state persistence", it seems:

Screenshot 2026-08-25 at 3 52 28 PM

What a bummer that it isn't included in the "hobby" version. 😅

dev-env/localstack/init-s3.sh
youseihuayu-wonderful: Added as suggested and cleared at startup.

ChengShi-1 (approved):
Looks good! Approving

PR #1065, opened 2026-09-09
Fix bootstrap CSS leaks when reusing Dataverse Design System

### LINKED ISSUE #1060: Reuse of design-system components leaks CSS into the consuming project

- When does this issue occur?
When reusing a design-system component in a different React-based project, which has `@iqss/dataverse-design-system` installed.

On pages of the reusing project that include any design-system component (tested with Button, Badge).

- What happens?
CSS side effects cause the background to become white (related to `--bs-body-bg: #fff`), and links to be styled differently (`color: var(--bs-link-color); text-decoration: underline;`).

- To whom does it occur (all users, curators, superusers)?
All users

No side effect, CSS only applies to the design-system component itself

`https://registry.npmjs.org/@iqss/dataverse-design-system/-/dataverse-design-system-2.2.0.tgz`

Not aware of any.

**Screenshots:**
Left side: A page where a `<Badge>` was included
Right side: The original page
Image

julian-schneider:
#1060 describes how reusing the Dataverse Design System leaked Bootstrap CSS into a consuming project. This PR is a proposal how this could be addressed:
- use `$prefix: dv-bs-` for Dataverse CSS
- stop bundling `bootstrap/scss/reboot` and `bootstrap/scss/type` into the default stylesheet, since they had major page-level effects
- introduce the optional `bootstrap-global.css` into the design-system dist, which does bundle `bootstrap/scss/reboot` and `bootstrap/scss/type`. It can be imported on demand, as an opt-in solution for those applications that want it.

- Closes #1060

1) Side-by-side comparison of `modern/` pages with and without these changes. If this PR is working correctly there should be no differences.
2) Use this version of the Dataverse Design System in a non-Bootstrap React project, and check if its CSS has any unwanted side effects. In project seen in the above linked issue's screenshot, no such issues occur anymore.

No

Users of the Dataverse Design System should be informed that importing the Bootstrap page-level styles is now opt-in (via `bootstrap-global.css`).

### DIFF
--- a/.storybook/preview.tsx
+++ b/.storybook/preview.tsx
@@ -1,5 +1,6 @@
 import React from 'react'
 import type { Preview } from '@storybook/react'
+import '@iqss/dataverse-design-system/bootstrap-global.css'
 import { ThemeProvider } from '@iqss/dataverse-design-system'
 import { createBrowserRouter, RouteObject, RouterProvider } from 'react-router-dom'
 import { FakerHelper } from '../tests/component/shared/FakerHelper'
--- a/packages/design-system/.storybook/preview.tsx
+++ b/packages/design-system/.storybook/preview.tsx
@@ -1,5 +1,6 @@
 import React from 'react'
 import type { Preview } from '@storybook/react'
+import '../src/lib/assets/styles/bootstrap-global.scss'
 import { ThemeProvider } from '../src/lib/components/theme/ThemeProvider'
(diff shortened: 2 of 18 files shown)


ChengShi-1:
Hi Julian, could you check the failure in component test and try to fix? (I think the failure from e2e test is not related and it just a flaky one)

PR #1066, opened 2026-09-09
fix: localize dynamic values and controlled labels (#1063)

### LINKED ISSUE #1063: Some dynamic values and search facets are not localized in the Modern Frontend

1. Change the Modern Frontend interface language to Spanish.
2. View a dataset containing an ingested tabular file.
3. Open the file details/metadata.
4. View search results with facets such as Subject, Author Name, and Author Affiliation.

**- When does this issue occur?**

When using the Modern Frontend in a non-English language. We observed the issue while testing the Spanish interface.

**- Which page(s) does it occur on?**

Dataset/file display pages
File metadata/details
Search results/facet sidebar

**- What happens?**

Some static UI text is translated correctly, but dynamically generated values and some Dataverse-controlled labels remain in English or use English-style formatting.

Examples observed:

1. File information
- File size is displayed as 180.3 KB instead of the Spanish locale format 180,3 KB.
- File type remains Tab-Delimited.
- The system-defined file tag Data remains in English.

Image

2. Search result dates

Dates displayed in search results also remain in English when the interface is set to Spanish. For example, an English-formatted date such as Sep 2, 2026 should instead be formatted according to the active Spanish locale.

Image

3. Search facets
- The Más... link is translated, but several facet headings remain in English, including:
- Subject
- Author Name
- Author Affiliation

4. Controlled Subject values also remain in English, for example:
- Social Sciences
- Arts and Humanities
- Medicine, Health and Life Sciences
- Earth and Environmental Sciences
- Law
User-supplied metadata such as author names and institutional affiliations should remain unchanged. The issue is with interface labels and Dataverse-controlled display values.

5. Facet result counts also do not appear to use locale-aware number formatting.

      For example, values are currently displayed as:
      53305
      4940
      
      In English, these would normally be formatted as:
      53,305
      4,940

      In Spanish, the formatting should follow the active Spanish locale, for example 53.305 or 53 305 and 4940 for es-ES, rather than using a fixed English or unformatted representation.

Image

**- To whom does it occur (all users, curators, superusers)?**

all users viewing the Modern Frontend

**- What did you expect to happen?**

When a supported interface language is selected, locale-sensitive values and Dataverse-controlled interface/display labels should consistently follow the active locale where translations are available.

For example:

English: 180.3 KB
Spanish: 180,3 KB

Search facet headings should also be localized, and controlled vocabulary labels such as Subject values should use their translated display labels where available.

Dataset version identifiers such as 4.0 or 3.1 should not be locale-formatted, since the period is part of the version identifier rather than a decimal separator.

v. 6.10.1 build iqss-4 | frontend version: 371cf9773

We have not identified an existing issue that covers these specific localization problems.

youseihuayu-wonderful:
- Format dates, file sizes, and counts using the active interface language.
- Localize known file types, system categories, facet headings, and Subject values.
- Preserve custom metadata, raw filter values, and dataset version identifiers.
- Prevent inherited object properties from being treated as translation keys.

- Closes #1063

Frontend-only changes; no backend API or dependency changes.

- Switch between English and Spanish and verify dates, sizes, and counts.
- Add and remove translated facets.
- Verify custom values, including constructor and toString, remain unchanged.
- Verify download sizes update after changing the interface language.

Validation: 121 passing tests across 16 affected specs; Node 22 build,
scoped ESLint, Prettier, and browser component smoke checks passed.
Full backend E2E was not run.

## Does this PR introduce a user interface change?
Yes: localized display values and labels; no layout changes.

## Is there a release notes or changelog update needed?
Updated CHANGELOG.md under Unreleased / Fixed.

## Additional documentation:
None.

### DIFF
--- a/tests/component/sections/collection/collection-items-panel/FacetsFilter.spec.tsx
+++ b/tests/component/sections/collection/collection-items-panel/FacetsFilter.spec.tsx
@@ -1,10 +1,15 @@
 import { FacetsFilters } from '@/sections/collection/collection-items-panel/filter-panel/facets-filters/FacetsFilters'
 import { CollectionItemsMother } from '@tests/component/collection/domain/models/CollectionItemsMother'
 import styles from '@/sections/collection/collection-items-panel/filter-panel/facets-filters/FacetsFilters.module.scss'
+import { CollectionItemsFacet } from '@/collection/domain/models/CollectionItemSubset'
+import i18n from '@/i18n'
 
 const facets = CollectionItemsMother.createItemsFacets()
 
 describe('FacetsFilters', () => {
+  beforeEach(() => cy.wrap(i18n.changeLanguage('en')))
+  afterEach(() => cy.wrap(i18n.changeLanguage('en')))
+
   it('should render skeleton while loading collection items and no facets', () => {
     cy.customMount(
       <FacetsFilters facets={[]} onFacetChange={cy.stub()} isLoadingCollectionItems={true} />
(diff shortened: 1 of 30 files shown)


youseihuayu-wonderful:
Hi Chen, I’ve opened a PR for #1063: https://github.com/IQSS/dataverse-frontend/pull/1066
 The fix covers locale-aware formatting and localized file/facet labels while preserving user-provided metadata and raw filter values. Local validation passed; GitHub CI was still running at my last check.
@ChengShi-1  Could you take a look when you are free?

ChengShi-1 (commented):
Confirmed in the UI with a `180,300-byte` file: English shows `180.3Kb`, while Spanish shows `180 milB` which should be `180,3KB`. Could we reuse `FileSize.toString(locale)` in `src/sections/collection/collection-items-panel/items-list/file-card/FileCardHelper.ts` `static formatBytesToCompactNumber`  so search results use proper file-size units consistently with the file details page? and create a test casue for it

tests/component/sections/collection/collection-items-panel/FacetsFilter.spec.tsx
```diff
@@ -1,10 +1,15 @@
 import { FacetsFilters } from '@/sections/collection/collection-items-panel/filter-panel/facets-filters/FacetsFilters'
 import { CollectionItemsMother } from '@tests/component/collection/domain/models/CollectionItemsMother'
 import styles from '@/sections/collection/collection-items-panel/filter-panel/facets-filters/FacetsFilters.module.scss'
+import { CollectionItemsFacet } from '@/collection/domain/models/CollectionItemSubset'
+import i18n from '@/i18n'
```
ChengShi-1, line 5: Could you move the `import i18n ...` to the beginning of the internal import group, after the external package imports, to keep the import order consistent with our usual style?

tests/component/sections/collection/collection-items-panel/file-card/FileCard.spec.tsx
ChengShi-1, line 1: I notice that we have more non-dynamic fields needed to be fixed and there are much more languages we haven't included, it's hard to test all of them in current approach even though it is one of the right methods. Thus, before extending this across the codebase, could we make the test setup more reusable?

- Centralize the default language reset and add a shared `cy.setLanguage()` helper, rather than repeating `i18n` imports and hooks in every spec.
- Test conversion, rounding, and locale formatting thoroughly in shared helper tests, using English and Spanish as representative locales.
- Keep component tests focused on integration: correct displayed values, updates after language switching, and preservation of raw API values.
- Add an explicit file-size assertion to the Spanish file-card test. The existing size assertion uses the production helper to calculate its expectation, so it can pass even when that helper produces the wrong format.

We don’t need to test every component in every language. Shared formatting tests plus focused component regression tests should give us good coverage without making the suite repetitive.

ChengShi-1 (commented):
Thanks for the fix! I left some comments below.

Also, I used AI (GPT) to scan the codebase and found additional areas where dynamic values and UI labels are not localized. My understanding is that this issue aims to improve Spanish localization more broadly, beyond the specific examples provided, so I think it would be worth addressing as many of these gaps as we reasonably can. That said, the scope appears to be much larger than originally expected, but AI could be really helpful for this kind of issue. I just created a sperate issue for this #1069

PR #1070, opened 2026-09-14
fix(IQSS#972): clear affiliation and description fields on collection…

### LINKED ISSUE #972: Edit Collection Page: clearing out affiliation field isn't working
Create a collection with an affiliation
Edit the collection by removing the affiliation
The Edit action succeeds, but the affiliation is still there

The affiliation should be removed from the collection

Latest development version

Janvi-kapoor:
### Which issue does this PR close?
Closes #972

### Summary of Changes
- When editing an existing collection, clearing out the `affiliation` or `description` input leaves `formData` as an empty string (`""`).
- Previously, this empty string was forwarded directly to `CollectionDTO`, causing the underlying update endpoint to ignore the field and retain the existing backend value.
- Sanitized `affiliation` and `description` in `useSubmitCollection.ts` so that empty/whitespace-only values resolve to `undefined`.
- Updated `CHANGELOG.md`.

### Testing
- Verified that clearing the input in the edit collection form submits `undefined` in the payload instead of `""`, allowing the backend to properly clear the stored value.


ChengShi-1:
@Janvi-kapoor  could you fix the lint error?

Janvi-kapoor:
> @Janvi-kapoor could you fix the lint error?

@ChengShi-1 i fix that can you re run workflow so that i can check is that fix is good or not

Janvi-kapoor:
hey @ChengShi-1  now all test passed you can review this pr.Thanks

ChengShi-1:
Hi @Janvi-kapoor , I can't reproduce it using current dev branch, I think this is fixed in the backend api. I am closing it since it is not repoduced. If you can reproduce, could you share a screen recording about how to reproduce this error in your end? so we could re-open this 

Create a collection with description and affiliation -> Save -> Edit description and affiliation to empty -> Save -> No  description and affiliation existing

